PR #145, opened 2019-07-24
Jonathan Patterson

Jonpatt92:
Here is my Pre-Work!


day_6/exercises/burrito.rb
```diff
@@ -5,15 +5,33 @@
...
-  attr_reader :protein, :base, :toppings
+  attr_accessor :protein, :base, :toppings
   def initialize(protein, base, toppings)
     @protein  = protein
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(toppings)
+    @toppings = toppings
```
timomitchel: This sets the instance variable to whatever you pass in, but we'd rather see you use an array method that adds the parameter that's passed in to the array. Can you find an Array method that can do that?

day_6/exercises/burrito.rb
```diff
@@ -5,15 +5,33 @@
...
+  end
+
+  def remove_topping
+    toppings.delete! toppings
```
timomitchel: Similarly, here we'd like to see you pass in a specific topping to remove, and then use an array method that deletes that topping from the `@toppings` array. Can you find one that can do that?

gear-up.md
```diff
@@ -0,0 +1,26 @@
+* What role does empathy play in your life and how has it helped you?
```
timomitchel: Hey Johnathan! We'd like to see you expand on these responses. We expect about a paragraph in response to each question. Please update these to be more robust. Your last response has the level of effort we are looking for.

career-dev.md
```diff
@@ -0,0 +1,29 @@
+*What is your greatest strength and how do you know?
```
timomitchel: Similar to the Gear Up, we'd like to see these responses expanded on. I encourage you to start telling your unique story, which you will certainly expand on and practice during your time at Turing.

day_6/exercises/burrito.rb
Jonpatt92: Honestly Tim I'm not sure what you mean by "an array method that adds the parameter that's passed into the array." or how to do that. It wasn't specified in the exercise that this was the expectation. 
I'm currently on the road/camping and won't be back until late Sunday evening. I don't think I can accommodate this change until I better understand what it is you're asking or how to do it.

day_6/exercises/burrito.rb
Jonpatt92: I couldn't find any documentation in the Day_6 lesson that speaks to what you're asking. I also couldn't find the info online or get any implementations to work. I'm going to be out of cell range and unable to process this further until late Sunday night. If you had gotten this to me sooner I would have more time to struggle through it.

PR #150, opened 2019-07-30
Zac Isaacson

zacisaacson:
Here is my prework.

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -11,9 +11,26 @@ class Burrito
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(topping)
+    @toppings << topping
+  end
+
+  def remove_topping(topping)
+    @toppings.delete_at(????)
+  end
+
+  def change_protein(protein)
+    @protein = protein
+  end
+
 end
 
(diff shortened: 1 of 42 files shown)


day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,26 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(topping)
+    @toppings << topping
+  end
+
+  def remove_topping(topping)
+    @toppings.delete_at(????)
```
timomitchel: Have not seen this syntax before `????`

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,28 @@
+numbers = {
```
timomitchel: Great job using this hash! This is what we like to see for this challenge!

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,14 @@
+puts "Please enter a number to count up to:"
```
timomitchel: Great job with fizzbuzz! I would recommend attempting all of the Day-7 challenges to build those ruby muscle!

PR #151, opened 2019-07-30
Devin Kapla

DekayHaHa:
Capstone Done

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -5,15 +5,32 @@
 # 3. change_protein
 
 class Burrito
-  attr_reader :protein, :base, :toppings
+  attr_accessor :protein, :base, :toppings
   def initialize(protein, base, toppings)
     @protein  = protein
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(t)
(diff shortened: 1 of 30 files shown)


day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,65 @@
+def in_words(int)
```
timomitchel: Great job using a hash here Devin! This is exactly the kind of implementation we were hoping to see. If you have time to attempt the other challenges this weekend then I encourage it! You can time box the challenges to get practice with a time constraint, which you will see weekly at Turing! Nice work on the technical capstone Devin!

day_6/exercises/burrito.rb
```diff
@@ -5,15 +5,32 @@
...
     @protein  = protein
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(t)
+    self.toppings.push(t)
+  end
+
+  def remove_topping(t)
```
timomitchel: Interesting implementation here! Sometimes - whatever works is whats best. But! ruby has a built in delete method as well that reads a little better. Good job on this exercise which combines some core Mod-1 skills.

PR #152, opened 2019-07-30
Tyler Balogh

t-balogh:
Here is my prework for capstone!

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -4,16 +4,16 @@
 # 2. remove_topping
 # 3. change_protein
 
-class Burrito
-  attr_reader :protein, :base, :toppings
-  def initialize(protein, base, toppings)
+class Burrito # => class
+  attr_reader :protein, :base, :toppings # => attributes
+  def initialize(protein, base, toppings) # => methods
     @protein  = protein
     @base     = base
     @toppings = toppings
   end
 end
 
-dinner = Burrito.new("Beans", "Rice", ["cheese", "salsa", "guacamole"])
-p dinner.protein
-p dinner.base
-p dinner.toppings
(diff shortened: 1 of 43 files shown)


day_6/exercises/burrito.rb
```diff
@@ -4,16 +4,16 @@
 # 2. remove_topping
 # 3. change_protein
 
-class Burrito
-  attr_reader :protein, :base, :toppings
-  def initialize(protein, base, toppings)
+class Burrito # => class
```
timomitchel: Hey Tyler! For this exercise we'd like to see you attempt to change the values that are initialized on line 16 with methods that exist within the class. Ruby has built in array methods that can handle both deleting and adding items to an array. You can also change the value of the protein with a method by setting the instance variable equal to whatever protein you pass into the method as an argument. You did a similar thing with the `speak` method in your `Classes_and_objects.rb` file. We'd like you to add some methods to this class that you can then call after you initialize the class on line 16. Let me know if you have questions about this!

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,66 @@
+#Ten little monkeys
```
timomitchel: Good job on this monkey challenge Tyler! Can you think of a way to DRY this up? That stands for `Don't repeat yourself`, and it is a common theme in programming. Any time you're repeating something, there is probably a way to do more by writing less. Maybe you could use a loop that executes 10 times? I like what you did with the array and using the index. See if you can refactor this to execute within some kind of loop, so that you have less lines of code.

day_7/checker_board.rb
```diff
@@ -0,0 +1,11 @@
+#checker board
```
timomitchel: Good job on this! I encourage you to attempt the rest of the Day-7 exercises to keep building those ruby muscles!

PR #153, opened 2019-07-30
Alex Gallant

agallant121:
Here is my Pre-work!

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -5,15 +5,40 @@
 # 3. change_protein
 
 class Burrito
-  attr_reader :protein, :base, :toppings
-  def initialize(protein, base, toppings)
-    @protein  = protein
-    @base     = base
-    @toppings = toppings
+  attr_accessor :protein, :base, :toppings
+
+  def initialize(protein: protein, base: base, toppings: toppings)
+      @protein  = protein
+      @base     = base
+      @toppings = toppings
+        puts "-" * 10
+    puts "These ingredients are available!"
+        puts "-" * 10
   end
+
+  def add_topping(topping)
+    @toppings << topping
+    @toppings
+
+  end
+
+  def remove_topping(topping)
+    index = @toppings.index(topping)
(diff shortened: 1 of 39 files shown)


day_6/exercises/burrito.rb
```diff
@@ -5,15 +5,40 @@
...
+
+  def add_topping(topping)
+    @toppings << topping
+    @toppings
+
+  end
+
+  def remove_topping(topping)
+    index = @toppings.index(topping)
+    @toppings.delete_at(index)
```
timomitchel: Nice job on this exercise Alex! As you may know, in ruby the last evaluated statement is returned from a method. Sometimes this is important if we are using that return value somewhere else in the program, but for the purposes of this exercise, once we change the @toppings array, we're not concerned with what happens after that. We also don't need to `p` each method call. We can just call the method and our code will run. Then you could `p` the dinner.toppings to see how the value changes after each method runs.

day_6/exercises/burrito.rb
```diff
@@ -5,15 +5,40 @@
...
+  end
+
 end
 
-dinner = Burrito.new("Beans", "Rice", ["cheese", "salsa", "guacamole"])
+dinner = Burrito.new(protein: "Beans", base: "Rice", toppings: ["cheese", "salsa", "guacamole"])
 p dinner.protein
 p dinner.base
-p dinner.toppings
+p dinner.toppings << ("onions") << ("peppers") << ("sour cream")
```
timomitchel: We want you to use the `add_topping` method to execute this functionality!

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,12 @@
+(1..100).each do |i|
```
timomitchel: Nice Fizzbuzz! I recommend trying your hand at all of the Day-7 exercises this weekend to keep building that Ruby muscle. You can time box them to 2 hours each for an extra challenge and to prepare for your weekly challenges at Turing.

PR #154, opened 2019-07-31
Brian Bower

BabsLabs:


day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,14 @@
+monkeys = ["Ten", "Nine", "Eight", "Seven", "Six", "Five", "Four", "Three", "Two"]
```
timomitchel: Great doing all of the Day-7 challenges Brian! You're looking strong coming in!

PR #155, opened 2019-07-31
Joshua O'Bannon 

jobannon:
Here is my Prework!


day_7/ten_little_monkeys.rb
```diff
@@ -0,0 +1,13 @@
+counter = 0
```
timomitchel: Hey Joshua! I love that your program accepts user input! Part of the challenge of this exercise is to try to print the string equivalent of the intengers. So instead of `1` have the program print `one` for every entry up to at least 10. Can you think of a data structure that can accomplish this?

PR #156, opened 2019-07-31
Ryan Hantak

rhantak:
Here is my prework!

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -11,9 +11,29 @@ class Burrito
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(new_topping)
+    toppings << new_topping
+  end
+
+  def remove_topping(topping_86)
+    @toppings.delete("#{topping_86}")
+  end
+
+  def change_protein(protein)
+    @protein = protein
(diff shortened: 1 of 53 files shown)


day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,29 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(new_topping)
+    toppings << new_topping
+  end
+
+  def remove_topping(topping_86)
+    @toppings.delete("#{topping_86}")
```
timomitchel: I think this 86 will work best without the string interpolation!

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,37 @@
+# Create a method for decided what to output for each number
```
timomitchel: Nice Fizzbuzz! I love that your accept user input and broke up your program into a method. If you have time this weekend then give a few of the other Day-7 challenges a go! You can time box them, so that you don't spend too much time on them, and you get some practice with timed challenges, which you will see weekly at Turing. Great job on the capstone Ryan!

PR #160, opened 2019-07-31
Laura Schulz

lrs8810:
Here is my prework!


day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,27 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
```
timomitchel: Nice work!

day_6/my_car.rb
```diff
@@ -0,0 +1,48 @@
+class MyCar
```
timomitchel: This is great work!

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,65 @@
+# Three little monkeys jumping on the bed,
```
timomitchel: Great job on these 2 day-7 challenges Laura! I love seeing that you broke up your concerns into a method and a class. If you have time to keep working this weekend, then I encourage you to attempt the other two challenges. You can time box these to get practice at working under a time constraint! You're looking strong coming into mod-1! Good luck!

PR #165, opened 2019-08-02
Gillings-Tyla-Mod-0-Capstone

tyladevon:
Prework complete!


day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,25 @@
+numbers = (1..100).to_a
```
timomitchel: Good job on Fizzbuzz! I would attempt all of the Day-7 challenges within the next few days before you start Turing. Struggling to come up with solutions to these exercises will pay dividends in Mod-1.

PR #167, opened 2019-08-04
Carleigh Crockett

Capleugh:
Here's my prework!


day_6/exercises/burrito.rb
```diff
@@ -5,15 +5,43 @@
...
 p dinner.protein
 p dinner.base
 p dinner.toppings
+dinner.add_topping
+dinner.remove_topping
+dinner.change_protein("chicken")
+dinner.protein = "chicken"
+p dinner.protein
+
+# I'm finding this particular exercise very confusing. I'm not sure whether I'm
```
timomitchel: Hey Carleigh! Thanks for the feedback and taking some time to talk out your thoughts here! So for this exercise we want you to add or remove whichever topping you pass into the `add_topping` or `remove_topping` methods. I see that you didn't pass anything in to either of them on this implementation, but the way you did it with `change_protein` is exactly what we want to see for that method. Ruby has some build in array methods that can handle the adding and removing of the topping within the method. Have you seen those before? If no this is a great chance to sharpen those googling skills! DM me if you have more questions about this.

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,32 @@
+f = (1..100).to_a
```
timomitchel: Nice fizzbuzz Carleigh! This is a solid implementation, because it works! Can you think of a way to break up the concerns into methods or a class? How about accepting user input? If you have time this weekend, then I encourage you to time box each of the other day-7 challenges to a few hours and give them a shot. This will be great practice for Mod-1 by giving yourself a time constraint, and it will ensure you don't spend too much time on them. Overall great job on the capstone and in Mod-0 Carleigh!

PR #172, opened 2019-08-09
Moduler_0_Capstone

sasloan:
Mod_0_capstone


day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,102 @@
+class Ryhme
```
timomitchel: Great job overall on your technical assessment Sebastian! I'm so encouraged by your progress in Mod-0. You're looking great coming in! Can you think of a way to make this class a bit more dry (don't repeat yourself)? Maybe one of the methods takes in the number of monkeys as a parameter and then uses interpolation? Maybe this is called within a loop? Just some suggestions to get you thinking. I encourage you to refactor this, and to attempt all of the day-7 exercises, so that your ruby muscles continue to get stronger. Remember you can always reach out with questions!

PR #173, opened 2019-08-10
Daniel Selph

danielpselph:


day_6/exercises/burrito.rb
```diff
@@ -5,12 +5,25 @@
...
 class Burrito
-  attr_reader :protein, :base, :toppings
+  attr_reader :protein, :base, :toppings)
   def initialize(protein, base, toppings)
     @protein  = protein
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(topping)
```
timomitchel: Hey Daniel! This is close to what we are looking for on this exercise, but unfortunately the += and -= syntax does not work for arrays. However! There are built in array methods that one can utilize in ruby that are not that far off what you have here. Can you find a method that will accomplish this in the docs? https://ruby-doc.org/core-2.4.1/Array.html

day_5/questions.md
```diff
@@ -2,12 +2,25 @@
...
 
+A hash is similar to array in the fact that it stores data however a hash allows for non-numerical data to be mapped
+
 1. In the space below, create a Hash stored to a variable named `pet_store`.  This hash should hold an inventory of items and the number of that item that you might find at a pet store.
 
+pet_store = {turtles: 100, goldfish: 200, rabbits: 50}
+
 1. given the following `states = {"CO" => "Colorado", "IA" => "Iowa", "OK" => "Oklahoma"}`, how would you access the value `"Iowa"`?
 
+puts "#{IA}"
```
timomitchel: Close! But this would actually be accessed by executing `states["IA"]`

day_5/questions.md
```diff
@@ -2,12 +2,25 @@
...
+p states.keys
+p states.values
+
 1. What is another example of when we might use a hash?  In this case, why is a hash better than an array?
 
+A hash would be helpful when mapping out all of the different names and details of a project you are creating. It would allow you to easily sort and change all of the different items you may have in your directory
+
 1. What questions do you still have about hashes?
+
+My questions would be how your hashes affect the code you are writing and how it exactly works with the other commands and methods we have been learning
```
timomitchel: Hashes have their own methods built in to ruby just like arrays, strings, and classes. Hashes are generally used for collections and always are organized into key/value pairs. Iterating over hashes is a common practice in programming in any language.

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,26 @@
+class Rhyme
```
timomitchel: This is a solid attempt Daniel! I'm glad that you broke this challene up into classes and methods, because that's what you will do with all of your projects in Mod-1. Can you think of a data structure that you could use that was a collection of integers that ll pointed to their string word equivalent? Part of the challenge here is getting the program to print words instead of integers, so 'one' instead of 1. Keep working on this and the other Day-7 challenges today. I recommend time boxing each challenge to about 1-2 hours so that you can practice with a time constraint, which you will see weekly in Mod-1.

day_6/questions.md
```diff
@@ -2,12 +2,30 @@
...
+cut_hair  
+remove_paw
+
 1. How do you create an instance of a class?
 
+Once your class has been created and initialized, you set the name of your instance equal to the class name followed by .new.
+
 1. What questions do you still have about classes in Ruby?
+
+My question about classes in Ruby is how to decide if your class is too general and when to break it down into multiple classes.
```
timomitchel: This is part of the art of programming. How one designs their program and separates concerns. It will generally change as your needs change. and you will learn some fundamental patterns of how to do this during your time at Turing. Some questions to keep in mind when designing. Is it easy to change my class? Does my class depend on other classes (hopefully not at all or very little). Does my class handle one concern or one chunk of functionality within my program (you want the answer to be yes).

Issue #171, opened 2019-08-05
Add video walkthrough of RBENV setup, installing a version of Ruby


josh-works:
I've got half of a video done for this - just a simple walkthrough of the "install RBENV and Ruby". 

I'll add to youtube, and pop a link into the setup instructions soon.

damwhit:
@josh-works , this is great! Can we make sure that we're having them use ruby version 2.5.1?

PR #174, opened 2019-08-29
Adding rbenv/ruby setup video, per #171,

josh-works:
Per #171, adding a few more videos. 

I need to edit/upload/add one more video before this is ready to go. I'll remove the WIP tag once that happens.


josh-works:
@damwhit, @timomitchel, I think this PR is ready to go. Just adds two short video walk-throughs, to close the gap between "finishing mod0" and "installing rbenv/ruby" + "forking/cloning this repository, doing the first bit of homework, adding/committing changes, and pushing it up to github."

I _think_ this will be the last block of video walkthroughs I do, at least until this has all sat out "in the wild", and we get feedback from students on potential gaps in the work. 

As always, please look it over and make any comments/feedback that pops to mind. I'll make changes as requested!

PR #253, opened 2020-01-11
Ashkan Abbasi

Ashkanthegreat:
Technical Capstone!

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -11,9 +11,23 @@ class Burrito
     @base     = base
     @toppings = toppings
   end
+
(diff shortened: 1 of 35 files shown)


day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,23 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(toppings)
+    puts "Add #{toppings}"
```
damwhit: For this and the two below methods, you're printing your arguments, but how would you leverage those arguments to update your instance variables?

PR #265, opened 2020-02-26
Norma Lopez

IamNorma:
Capstone

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -11,9 +11,27 @@ class Burrito
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(toppings)
+    @toppings = toppings
+    p "You added: #{toppings}"
+  end
+
+  def remove_topping(toppings)
+    p "You removed: #{toppings}"
+  end
+
+  def change_protein(protein)
+    @protein = protein
+  end
 end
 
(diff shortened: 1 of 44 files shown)


day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,27 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(toppings)
+    @toppings = toppings
```
damwhit: Right now you're reassigning your toppings, how would you change it so that you're adding to the existing array?

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,27 @@ def initialize(protein, base, toppings)
...
+    p "You added: #{toppings}"
+  end
+
+  def remove_topping(toppings)
+    p "You removed: #{toppings}"
```
damwhit: How would you change it so that you're removing and element from the existing array?

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,27 @@ def initialize(protein, base, toppings)
...
+  end
+
+  def change_protein(protein)
+    @protein = protein
```
damwhit: nice work here!

PR #267, opened 2020-02-26
Danny Ramos

muuuydanny:


day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,27 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(topping)
+    @toppings = toppings << topping
+  end
+
+  def remove_topping(topping)
```
damwhit: How would you change this method so that it is removing a topping from the ingredients array?

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,27 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(topping)
+    @toppings = toppings << topping
```
damwhit: how could you simplify this statement to accomplish the same thing?

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,23 @@
+`10 LITTLE MONKEYS`
+
+for i in (10).downto(1)
```
damwhit: Enumerables in ruby are typically preferred to for loops. How might you change this to use `.each` instead of a for loop?

PR #271, opened 2020-02-28
Judith Pillado

judithpillado:
Here is my pre-work!


day_6/exercises/burrito.rb
```diff
@@ -1,5 +1,5 @@
 # Add the following methods to this burrito class and call the methods below
-#the class:
+# the class:
```
damwhit: How would you add the following methods to the first class below? Try your best to update the @toppings and @protein instance variables accordingly.

PR #281, opened 2020-03-04
James Belta

JBelta:
Prework!

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -13,7 +13,27 @@ class Burrito
   end
 end
 
+def add_topping(topping)
+  p "Can you add #{topping} to my burrito?"
+end
+
+def remove_topping(topping)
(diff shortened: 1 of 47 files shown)


day_6/exercises/burrito.rb
```diff
@@ -13,7 +13,27 @@ def initialize(protein, base, toppings)
   end
 end
 
+def add_topping(topping)
```
damwhit: @JBelta these three methods are a good start. But I'd expect them to be methods defined within the Burrito class. Also, how could you utilize the methods to update the instance variables of a given burrito object?

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,34 @@
+numbers = [1]
+
+numbers.map {|n| numbers << n + 1
```
damwhit: How would using .each change the output of your code? Why use .map over .each in the case?

PR #294, opened 2020-04-05
Joshua Carey

jdcarey128:
Mod0 capstone


day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,11 @@
+monkeys = ["One", "Two", "Three", "Four", "Five", "Six", "Seven", "Eight", "Nine", "Ten"]
+
+monkeys.reverse.each do |monkey|
```
damwhit, line 3: Nice work here, utilizing reverse!

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,13 @@
+numbers = (1..100)
+
+numbers.each do |number|
```
damwhit, line 3: Very clean! Good syntax and style!

PR #304, opened 2020-04-13
Nick Edwin

NickEdwin:
Here is my prework!

### DIFF
--- a/day_1/exercises/ex1.rb
+++ b/day_1/exercises/ex1.rb
@@ -1,0 +1,11 @@
+# puts "Hello World!"
+# puts "Hello Again"
+# puts "I like typing this."
+# puts "This is fun."
+# puts "Yay! Printing."
+# puts "I'd much rather you 'not'."
+# puts 'I "said" do not touch this.'
+
+puts "My nme is Nick Edwin."
+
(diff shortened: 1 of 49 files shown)


robertguturing:
@NickEdwin thanks for you hard work on this. I left some feedback below that you can think over or even address to continue your learning and growth mindset.

* Naming in programming is really important, on[ this line of code ](https://github.com/NickEdwin/backend_module_0_capstone/blob/master/day_7/10_little_monkeys.rb), you chose to name the variable poem.  Variables assigned to an array are generally pluralized and reveal something about the information it contains.  So for example if I had an array of instructor names, it would look something like this.

`instructor_names = ["Ellen", "Bob", "David"]`

What would you name this variable instead knowing that???

* Your use of the enumerable `select` [here is](https://github.com/NickEdwin/backend_module_0_capstone/blob/master/day_7/10_little_monkeys.rb) misleading. When you simply want to print something to the screen you should use the `each` enumerable method.  `select' works like this.

```
[1, 2].select do |number|
  number.even?
end

======> This will return an array [2] as it is the only even number. You have to return a boolean within the codeblock and anything that returns `true` will be added to the result array.  
```

PR #305, opened 2020-04-13
Jonathan Wilson

Jonathan-M-Wilson:
Here is my prework!

### DIFF
--- a/day_1/exercises/ex1.rb
+++ b/day_1/exercises/ex1.rb
@@ -1,0 +1,16 @@
+puts "Hello World!"
+puts "Hello Again"
+puts "I like typing this."
+puts "This is fun."
+puts "Yay! Printing."
+puts "I'd much rather you 'not'."
+puts 'I "said" do not touch this.'
+
+#study drill 1
+puts "Another line is right here."
+
+#study drill 2 completed. Put "#" in front of every line but one.
+
+#study drill 3. "#" turns that line into a comment 
+
+# puts "Notes for days."
--- a/day_1/exercises/ex2.rb
+++ b/day_1/exercises/ex2.rb
@@ -1,0 +1,15 @@
+# A comment, this is so you can read your program later.
+# Anything after the # is ignored by ruby.
+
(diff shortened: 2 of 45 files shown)


robertguturing:
@Jonathan-M-Wilson thanks for your hard work on this. I left some feedback and answers to your questions on here. No need to address the feedback but certainly look into it if you have time.

* I left feedback on your 10 little monkeys exercise in this slack thread. https://turingschool.slack.com/archives/G010Q1WMWP7/p1587487858131600

* The `(1..whatever_number)` i thought at first created an array. But I was wrong! After trying it out in `irb` it looks like that snippet of code creates a `Range` object.  Without digging too much into it, it looks like you can call enumerable methods on range objects, which allows you to run methods like `each` and `reverse_each`.  Today I learned something.

* Your fizzbuzz code is clean for the most part, but I have a small nitpick. I think in this case it would be better first to check if a number is divisible by 3 and 5 in your first if statement. This would allow you to remove the second part of this conditional you have in your code `(number % 3 == 0) && !(number % 5 == 0)`

So instead of
```
 if (number % 3 == 0) && !(number % 5 == 0)
      puts 'Fizz'
    elsif (number % 5 == 0) && !(number % 3 == 0)
      puts 'Buzz'
    elsif (number % 3 == 0) && (number % 5 == 0)
        puts 'FizzBuzz'
    else
      puts number
    end
```

You can do
```
if (number % 3 == 0) && (number % 5 == 0)
      puts 'FizzBuzz'
    elsif (number % 5 == 0)
      puts 'Buzz'
    elsif (number % 3 == 0)
        puts 'Fizz'
    else
      puts number
    end
```

Another solution is you could simply build a string object that you print out at the end of the conditional so.

```
string_to_print = ''

    if (number % 3 == 0)
      string_to_print += 'Fizz'
    elsif (number % 5 == 0)
       strint_to_print += 'Buzz'
    else
       strint_to_print += number.to_s
    end

    puts string_to_print
```

PR #308, opened 2020-04-15
Update rbenv install instructions to use zsh

josh-works:
MacOS defaults to `zsh`, not `bash`; the instructions here listed `bashrc` specific commands. 

It's caused some problems for students recently, so I'd like to avoid sabotaging even _more_ students.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ $ rbenv init
 .
 .
 # Load rbenv automatically by appending
-# the following to ~/.bash_profile:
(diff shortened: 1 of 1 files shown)


damwhit:
@josh-works thanks for taking the time to do this!

josh-works:
> thanks for taking the time to do this!

Well, I was the person that wrote the thing that was now out-of-date.

I hate seeing my name come up when someone angrily `git blame`'s something. 😬

PR #310, opened 2020-04-15
Michael Chickillo

Truthvendor:
Here is my pre-work.

### DIFF
--- a/day_1/README.md
+++ b/day_1/README.md
@@ -1,4 +1,4 @@
-# Day 1 - Strings and Numbers
+``# Day 1 - Strings and Numbers
 
 Today, you will begin to learn the basics of Ruby. You will work through several tutorials which will help build your comfort running files and seeing output. You will also learn about two ways to represent data in Ruby: *Strings* and *Numbers*.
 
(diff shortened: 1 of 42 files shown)


Truthvendor:
Gear Up link:
https://gist.github.com/Truthvendor/6e86989003fe7feaf2e2c010be885ed8

Professional Development link:
https://gist.github.com/Truthvendor/ccdc737fdc72de91cf502fb41c87fbb8

Technical Pre-work link:
https://github.com/Truthvendor/backend_module_0_capstone

robertguturing:
@Truthvendor thanks for your hard work on this. I left some feedback below that I think would be really helpful for you to address to enhance your knowledge. 

* Your fizzbuzz example is actually not working correctly.  I copied and pasted the code into irb and when a number is divisible by 3 and 5 the word FizzBuzz never appears on one line.  Can you recheck your code here? Hint it has something to do with the way your conditional statements are ordered.

* Although the 10 little monkeys solution is correct.  Can you do it again but use some sort of looping construct similar to what you did in your fizzbuzz example?? Whenever we have multiple items we want to iterate over creating a loop is the tool we want to reach for to accomplish that solution.

PR #311, opened 2020-04-15
Ryan Laleh

RyN21:


robertguturing:
@RyN21 thanks for your hard work on your capstone. I left some feedback below that would be great if you took a look at it. Thanks!

* On these lines of code it looks like you are assigning attr_accessor attirbutes and attr_reader attributes to the same attributes.  This is redundant. Could you go research what `attr_reader, attr_accessor, and attr_writer` do and tell me how they are different? Then could you fix the code to reflect your new learnings?

PR #314, opened 2020-04-15
Michael Evans

mjevans113:
Mod 0 Back-End Capstone


robertguturing:
@mjevans113 Thanks for your hard work on this. I left some feedback below.

* Great job on your 10 little monkeys exercise I put some feedback i this slack thread - https://turingschool.slack.com/archives/G010Q1WMWP7/p1587490397133100

* In your fizzbuzz example there is a way to solve it without using nested conditionals.  A nested conditional looks like this
```
if #something
  if #something
  else
  end
end
```

PR #318, opened 2020-05-18
Jose Lopez

JoseLopez235:


day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,14 @@
...
+monkeys = gets.chomp.to_i
+
+def monkey(monkeys)
+  if monkeys != 0
+    puts  "#{monkeys} little monkeys jumping on the bed,
+    One fell off and bumped his head,
+    Mama called the doctor and the doctor said,
+    'No more monkeys jumping on the bed!'"
+    puts ""
+    monkey(monkeys - 1)
```
damwhit, line 11: Whoa I see you using some recursion. So fancy

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,16 @@
+puts "Give me any number: "
+number = gets.chomp.to_i
+puts ""
+
+number.times do |i|
+  i += 1
```
damwhit, line 6: This accumulator is redundant I believe

day_7/10_little_monkeys.rb
damwhit: An iterative approach would likely be simpler

day_6/exercises/object_model.rb
```diff
@@ -0,0 +1,70 @@
...
+  attr_reader :year
+  def initialize(year, model, color)
+    @year = year
+    @model = model
+    @color = color
+    @current_speed = 0
+  end
+
+  def speed_up(number)
+    @current_speed += number
```
damwhit, line 32: Nice use of a parameter here.

day_7/caesar_cipher.rb
```diff
@@ -0,0 +1,34 @@
+class CeasarCipher
+  def initialize(words, number)
+    @words = words
+    @number = number
+    @alphabet = ["A","B","C","D","E","F","G","H","I","J","K","L","M","N","O","P","Q","R","S","T","U","V","W","X","Y","Z"]
+  end
+
+  def encoded_alphabet
```
damwhit, line 8: Prefer action like present tense

JoseLopez235:
Thank you for the feedback. So there are no changes that I need to make?

day_7/fizzbuzz.rb
JoseLopez235: The reason I made that accumulator is that if I didn't it will always be short by 1. For example if I gave it a number of 20 the program will only go up to 19 and stop there. That's the reason I add 1 to every number that enters the block, for it can get to 20.

PR #319, opened 2020-05-20
Christopher Allbritton

Callbritton:
Here's my prework :)


day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,18 @@
+def fizzbuzz
+  (1..100).map do |n|
```
damwhit, line 2: Nice work here! Is there a reason why you used .map instead of .each?

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,34 @@
...
+  4 => 'Four',
+  5 => 'Five',
+  6 => 'Six',
+  7 => 'Seven',
+  8 => 'Eight',
+  9 => 'Nine',
+  10 => 'Ten'
+}
+
+num.reverse_each do |x, y|
```
damwhit, line 14: What are x and y referring to here? Can you think of more descriptive names?

day_6/exercises/burrito.rb
```diff
@@ -1,19 +1,35 @@
...
+  def change_protein(protein)
+    @protein = protein
+  end
 end
 
 dinner = Burrito.new("Beans", "Rice", ["cheese", "salsa", "guacamole"])
 p dinner.protein
 p dinner.base
 p dinner.toppings
+p dinner.add_topping(["sour cream"])
```
damwhit, line 33: How could you rewrite the method above so that you can pass a string instead of an array here?

day_7/fizzbuzz.rb
Callbritton: .each returns the original unchanged object (1..100) and, in the context of the way this code is written returns the entirety of the FizzBuzz strings as desired. .each could be utilized, but .map has the desired functionality and is cleaner.

day_7/10_little_monkeys.rb
Callbritton: Hi!

x: key
y: value 

key/value would be better variable names (I have updated the file accordingly). monkey_int/monkey_string could work as well!

day_6/exercises/burrito.rb
Callbritton: I think you are talking about line 33. If so, you could implement the shovel operator on the string "sour cream":

`p dinner.add_topping(["sour cream"])`
to: 
`p dinner.toppings << "sour cream"`

I updated this file with this change as well :)

PR #320, opened 2020-05-21
Kevin Cuadros

kevxo:
Here is my prework!

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -1,4 +1,4 @@
-# Add the following methods to this burrito class and 
+# Add the following methods to this burrito class and
 # call the methods below the class:
 # 1. add_topping
 # 2. remove_topping
@@ -11,9 +11,21 @@ class Burrito
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping
+  end
+
+  def remove_topping
+  end
+
+  def change_protein
+  end
 end
 
 dinner = Burrito.new("Beans", "Rice", ["cheese", "salsa", "guacamole"])
 p dinner.protein
 p dinner.base
 p dinner.toppings
+p dinner.add_topping
+p dinner.remove_topping
+p dinner.change_protein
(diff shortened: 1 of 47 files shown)


day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,12 @@
+monkey = ["Ten", "Nine", "Eight", "Seven", "Six", "Five","Four", "Three", "Two", "One"]
```
damwhit, line 1: Pluralization is important, when naming! This is multiples so I'd reflect that.

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,12 @@
...
+monkey.each do |ele|
```
damwhit, line 2: Then here you could do `monkeys.each do |monkey|` which I think would be more readable

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,12 @@
...
+  1.times {p "#{ele} little monkey(s) jumping on the bed, One fell off and bumped his head, Mama called the doctor and the doctor said,  'No more monkeys jumping on the bed!'"}
```
damwhit, line 3: what happens if you take away the times block here?

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,28 @@
...
+    puts ele
+  end
+end
+
+
+#bonus
+puts "Enter any number"
+num = gets.chomp.to_i
+
+num.times do |ele|
```
damwhit, line 18: what would be a better name for this block variable?

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,21 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping
```
damwhit, line 15: @kevxo how would you update the data model in this method and the two below?

day_7/10_little_monkeys.rb
kevxo: Yea that makes sense

day_7/10_little_monkeys.rb
kevxo: Ok. Thanks.

day_7/10_little_monkeys.rb
kevxo: It will still work fine and now I believe there was no need of the times block.

day_7/fizzbuzz.rb
kevxo: number

day_6/exercises/burrito.rb
kevxo: I just updated everything through the Problem Fix commit and Error Commit

PR #322, opened 2020-05-26
Angela Guardia

AngelaGuardia:


day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,28 @@
...
+  8 => "Eight",
+  9 => "Nine",
+  10 => "Ten"
+}
+
+# an array in reverse for rhyme
+arr = (1..number).to_a.reverse
+
+# print each rhyme
+arr.each do |i|
```
damwhit, line 24: The nursery rhyme does change on the last line. How would you update your code to account for that?

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+  elsif i % 5 == 0 # if multiple of 5
+    print "Buzz, "
+  else
+    print i.to_s + ", " # if none
+  end
+end
+
+# Questions
+# I wasn't sure whether nesting if statements is bad practice? I was trying to
```
damwhit, line 22: It's not ideal. In this case how could you change it to use an && statement instead?

day_6/exercises/good_dog.rb
```diff
@@ -0,0 +1,23 @@
+class GoodDog
+  attr_accessor :name, :height, :weight
+
+  def intialize(n, h, w)
+    @name = n
```
damwhit, line 5: Always avoid using 1 word var or parameter names. It's best to be clear and air on the side of verbosity.

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,27 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(topping)
+    @toppings << topping
```
damwhit, line 16: Great use of pluralization/lack there of here

damwhit (commented):
@AG2006 nice work on this! Keep it up moving forward!

PR #327, opened 2020-06-02
Nell Smart

secretbookclub:
Here is my backend Capstone!

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -1,19 +1,48 @@
-# Add the following methods to this burrito class and 
+# Add the following methods to this burrito class and
 # call the methods below the class:
 # 1. add_topping
 # 2. remove_topping
 # 3. change_protein
 
 class Burrito
-  attr_reader :protein, :base, :toppings
+  #attr_accessor :protein, :base, :toppings
+  attr_accessor :protein, :base, :toppings
+
   def initialize(protein, base, toppings)
-    @protein  = protein
-    @base     = base
+    @protein = protein
+    @base = base
     @toppings = toppings
+  end
+
+  def info # kept trying to define this method with the variables after the method name, figured out that doesn't work.
+    p "My burrito has #{self.protein}, #{self.base}, and #{self.toppings}."
+  end
+
+  def add_topping(toppings)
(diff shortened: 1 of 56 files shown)


day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,68 @@
+# I explored a few different solutions...
+
+# 1.
+
+10.downto(2) do |number|
```
damwhit, line 5: I like this solution, because it is simplest.

damwhit: Love that you explored multiple!

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,17 @@
...
+    puts "fizzbuzz" #elsif?
+
+  elsif i % 3 == 0
+    puts "fizz"
+
+  elsif i % 5 == 0
+    puts "buzz"
+
+  else puts i # It also worked to end the conditional after the last elsif
+# statement, then put the else statement right before the end of the iteration
```
damwhit, line 14: love that you're playing around with it. Why do you think?

day_6/exercises/dog.rb
```diff
@@ -1,11 +1,11 @@
 # In the dog class below, write a `play` method that makes
-# the dog hungry. Call that method below the class, and 
+# the dog hungry. Call that method below the class, and
 # print the dog's hunger status.
 
 class Dog
   attr_reader :breed, :name, :age, :hungry
 
-  def initialize(breed, name, age)
+  def initialize(breed, name, age) # why don't we need to include hungry in these variables when we are defining it below as an instance variable? My guess - Boolean values are an exception
```
damwhit, line 8: Good question. I'd probably add it here as well

day_6/exercises/burrito.rb
```diff
@@ -1,19 +1,48 @@
...
+    @protein = protein
+    @base = base
+    @toppings = toppings
+  end
+
+  def info # kept trying to define this method with the variables after the method name, figured out that doesn't work.
+    p "My burrito has #{self.protein}, #{self.base}, and #{self.toppings}."
+  end
+
+  def add_topping(toppings)
```
damwhit, line 21: I would only expect this to add a single topping. How would you change this and calling the method below?

PR #328, opened 2020-06-03
Luke Hunter James-Erickson

LHJE:
Here is my prework!


day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,8 @@
+1.upto(100) do |n|
+  print "Fizz" if a = (n % 3).zero?
```
damwhit, line 2: Missing a case here

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,28 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(new_topping)
+    toppings.push(new_topping)
+  end
+
+  def remove_topping(old_topping)
```
damwhit, line 19: What would be a better name for the parameter here?

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,48 @@
+class MonkeyBed
+  attr_accessor :monkeys, :plural, :last_sentence
```
damwhit, line 2: I do think that this is a pretty unique solution. Nice work. However, I'd recommend trying out a simpler solution with a loop. Should only require one method definition.

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,8 @@
...
+  print "Buzz" if b = (n % 5).zero?
+  print n unless (a || b)
+  puts
+end
+
+# I'm gonna be honest here.  I found this one online after trying for over an hour to figure out something related to (1..100) and `if`/`elsif` and .join(', '), and because I'm working full time up until Turing starts, I wanted to get this in before the deadline.  I know that copying and pasting from the internet is, I guess, part of this whole thing, but it's not what I want to be doing because it's not teaching me what I need. I will be reading through the books and grinding what even is `.zero?` doing in this, but I wanted to put it out there that, yeah, I can tell that this is a bit advanced for where we were.
```
damwhit, line 8: @LHJE I'd like for you to redo this. Try to pseudocode it out beforehand. It should require iteration over a collection of numbers and 3 conditionals + an else.

day_6/exercises/burrito.rb
LHJE: Sorry, I clicked "Resolve conversation" thinking it would take me to what you were asking me to resolve.  I'll look at it and click "Resolve conversation" when it's actually resolved!

PR #329, opened 2020-06-03
Tyler Fields

fieldstyler:
Here is my prework!


day_6/exercises/person.rb
```diff
@@ -1,5 +1,69 @@
...
+class Person
+  attr_reader :hair_color, :skin_color, :height, :weight
+  def initialize(hair_color, skin_color, height, weight)
+    @hair_color = hair_color
+    @skin_color = skin_color
+    @height = height
+    @weight = weight
+  end
+
+  def dye_hair=(c)
```
damwhit, line 15: always air on the side of verbosity and clarity. So I would name the parameter here `color` instead of `c`

day_6/exercises/burrito.rb
```diff
@@ -1,19 +1,55 @@
...
-p dinner.toppings
+dinner = Burrito.new("Bean", "Rice", ["cheese", "salsa", "guacamole"])
+print puts "-" * 15
+puts dinner.protein
+puts dinner.base
+puts dinner.toppings
+puts "-" * 15
+print dinner.combo
+puts "-" * 15
+dinner.change_protein = "Chicken"
```
damwhit, line 43: Note that you're using an assignment operator here. How would you pass this info as an argument?

day_6/exercises/burrito.rb
```diff
@@ -1,19 +1,55 @@
...
-  attr_reader :protein, :base, :toppings
+  attr_accessor :protein, :base, :toppings
   def initialize(protein, base, toppings)
     @protein  = protein
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping=(toppings)
+    @toppings = @toppings << toppings
```
damwhit, line 16: If you're only adding a single topping, what would be a better name for your parameter here?

PR #331, opened 2020-06-05
Travis McKinstry

TravisGM92:


francepack:
@TravisGM92 
Well done with your capstone- I liked seeing all of the extra practice, as well as effort given to each of the day 7 problems. All the extra practice will help you feel that much more comfortable in Mod1! I wanted to provide a little feedback:

- I ran your day7 cipher. Javascript is my primary language (former FE student), so ruby methods like .scan and .rotate stand out to me... ruby has quite a few more baked-in methods than js! Your solution didn't quite work the way I expected, but I like how you have set it up. I think within your .each logic, you will need to make sure letters are delivered back in the correct order, and if a new index goes beyond the alphabet set, that this is accounted for. When I logged `user_index`, that seemed to produce funky results sometimes.

francepack:
@TravisGM92 
And, I believe your cipher does actually work! Not sure what I was seeing before.
I tried some different test cases, and it was working well. Nice job!

TravisGM92:
@francepack Sweetness! Thanks for checking that. I'll play around with it more to make sure it works.

PR #333, opened 2020-06-07
Roberto Basulto

Eternal-Flame085:


day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,15 @@
+def fizzbuzz(num)
+  count = 1
+  while count <= num do
```
damwhit, line 3: .each is more canonical in ruby. How would you implement that instead?

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,29 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(topping)
```
damwhit, line 15: Good use of singular vs. plural naming here. It's clearer to me reading it.

PR #336, opened 2020-06-09
Corey Carter

crc2231:
Corey Carter's Pre-Work!


francepack:
@crc2231 
Well done overall with your capstone! All of this extra practice will help you feel that much more comfortable in Mod1. I wanted to provide a little feedback:

- I liked seeing the different uses of array methods within your class methods. I think building classes is a great way to keep up more experimentation.

- Right now, your checkerboard prints lines that are the same length. A next step would be to consider how you might vary the length of each line, like if the size was 3, each line would get something like: 
```
 X 
X X
 X 
```

- I encourage you to try the cipher! These problems are great exercises in pseudocoding the steps you expect to take, then trying out some code.

PR #337, opened 2020-06-10
Daniel Lessenden

D-Lessenden:
Backend Mod 0

### DIFF
--- a/day_0/README.md
+++ b/day_0/README.md
@@ -60,7 +60,7 @@ Don't worry about putting any text into these files. For now, just create this s
 
 ### The rm Command
 
-Be careful when deleting something though the terminal! While we want to be comfortable using the very helpful `rm` command, once executed, the command cannot be undone. Lets learn more about `rm` before we practice:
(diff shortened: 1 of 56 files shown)


francepack:
@D-Lessenden 
Well done with your capstone- I liked seeing all of the extra practice, as well as effort given to each of the day 7 problems. All the extra practice will help you feel that much more comfortable in Mod1! I wanted to provide a little feedback:

- Regarding your question about adding to hashes, I most frequently find myself just adding new key/value pairs individually via dot notation.

- I really like seeing how you break apart the checker board problem. I encourage you to write out your thoughts in that style for coding problems you'll face!

- Keep up experimentation with ruby methods! Building classes is a great way to try making methods that utilize some different methods, allowing an opportunity to see what happens when you use them.

Nice work! Keep up the momentum until Turing starts!

francepack:
@D-Lessenden 
I realized after the fact I left some 'opinionated' feedback on accident- I think its more common to use bracket notation to add key/value pairs in ruby. (dot notation with key/value pairs is more of a javascript notation).

PR #338, opened 2020-06-10
HOPE GOCHNOUR

hopesgit:
One serving of pre-work, ready for evaluation!

### DIFF
--- a/day_1/exercises/ex1.rb
+++ b/day_1/exercises/ex1.rb
@@ -1,0 +1,9 @@
+puts "Hello World!"
+# puts "Hello Again!"
+# puts "I like typing this."
+# puts "This is fun."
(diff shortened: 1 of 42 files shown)


francepack:
@hopesgit 
Well done with your capstone! All of this extra practice will help you feel that much more comfortable in Mod1! I wanted to provide a little feedback:

- I noticed in some of your classes, you wrote your methods in a separate block from the classes attributes. Typically, attributes and methods will be within the same `class` block.

- I love seeing an OOP cipher solution! It looks like you did a lot of research to break apart that problem, and found a great way to solve it. Just remember to spend time digesting the code line by line to be able to explain it all, and it never hurts to practice alternate strategies for any of those problems for practice!

PR #339, opened 2020-06-10
Shaun James

ShaunDaneJames:
Here is my Mod 0 work! I tried a few variations at the end. I had a feeling OOP was supposed to be the solution; however, I tried my best to solve it with and without OOP. PLEASE give me feed back, let me fix my mistakes (cause I know I made few, or maybe many)! I feel that the feedback on the home works were a little disconnected with the progress on the capstone work.

Gear up:
https://gist.github.com/ShaunDaneJames/44dfc4e4efd03644787dcca2f5ad98e5

Parin results:
https://gist.github.com/ShaunDaneJames/b89f98549afd116f3a0a1487ffd0a53c

Career Journal:
https://gist.github.com/ShaunDaneJames/0de0e9c2461077029f14e28388017c40

Capstone:
https://github.com/ShaunDaneJames/backend_module_0_capstone

Thanks!
-Shaun

### DIFF
--- a/day_7/checker_board_2.rb
+++ b/day_7/checker_board_2.rb
@@ -1,0 +1,14 @@
+puts "Please indicate a size for your checker board:"
+
+size = gets.chomp.to_i
+
+puts "```\n"
+
+for num in 1..size do
+  if num % 2 != 0
+    puts "X X X"
+  else puts " X X X"
+  end
+end
+
+puts "```"


francepack:
@ShaunDaneJames 
Well done with your capstone- I liked seeing all of the extra practice. All the extra practice will help you feel that much more comfortable in Mod1. I wanted to provide a little feedback:

- I really liked that you took it upon yourself to craft multiple approaches to solutions! I'd say the best approach to 10 Little Monkeys involves iteration, and getting an input from a user, but the class approach is quite unique. Also, iteration should be involved with any fizzbuzz answer, but I think you could incorporate that into your other approaches. I really like the creativity at this point, and willingness to try different strategies!

- I'd encourage you to try the cipher and checkerboard! You've got a good start with your 'high-level' thinking. Try to translate that into specific pseudocode-like steps. For instance, how will you use an alphabet array for reference? Or how will you use an input to construct a line of Xs? Challenging yourself with these is great Mod1 practice, even if totally working solutions don't happen at this point.

PR #340, opened 2020-06-10
Garret Wolf

garretwolf:


day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,27 @@
...
+  'Seven',
+  'Six',
+  'Five',
+  'Four',
+  'Three',
+  'Two',
+  'One'
+]
+
+monkeys.each do |monkeys|
```
mschae16, line 16: @garretwolf - Nice work here! One suggestion is to change the placeholder name of your current element in the iteration to something _singular_ and _different_ from the name of the array, ie. line 15 could become
`monkeys.each do |monkey|`

Keep it up!

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,19 @@
+def fizzbuzz(minimum, maximum)
```
mschae16, line 1: Excellent work!

PR #341, opened 2020-06-10
Elliott Stein

estein1988:
Here is my completed Mod 1 (BE) capstone!

### DIFF
--- a/day_7/10_little_monkeys.rb
+++ b/day_7/10_little_monkeys.rb
@@ -32,7 +32,7 @@ while n >= 1
     He fell off and bumped his head,
     Mama called the doctor and the doctor said,
     \"Get those monkeys right to bed!\""
-  else n <= 9999999999
+  else
     puts "
     #{n} little monkeys jumping on the bed,
     One fell off and bumped his head,


day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,45 @@
+plural = ['Ten', 'Nine', 'Eight', 'Seven', 'Six', 'Five', 'Four', 'Three', 'Two']
+single = ['One']
+base = "jumping on the bed,
+One fell off and bumped his head,
+Mama called the doctor and the doctor said,
+\"No more monkeys jumping on the bed!\""
+
+plural.each do |words|
```
mschae16, line 8: @estein1988 - Nice work on iterating through the arrays here to print the appropriate rhyme! For line 8, I think it would make more sense to name the current item being iterated over (`|words|`) something singular such as `word`, `number`, etc.

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,45 @@
...
+n = $stdin.gets.chomp.to_i
+
+while n >= 1
+  if n == 1
+    puts "
+    1 little monkey jumping on the bed,
+    He fell off and bumped his head,
+    Mama called the doctor and the doctor said,
+    \"Get those monkeys right to bed!\""
+  elsif n <= 9999999999
```
mschae16, line 35: I think line 35 could just be an `else` statement rather than an `elsif`

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,45 @@
...
+    puts "
+    #{n} little monkeys jumping on the bed,
+    One fell off and bumped his head,
+    Mama called the doctor and the doctor said,
+    \"No more monkeys jumping on the bed!\""
+  else
+    puts n
```
mschae16, line 42: What's the purpose of line 42?

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,40 @@
...
+while x <= 100
+  if x % 3 == 0
+    puts "Fizz"
+  elsif x % 5 == 0
+    puts "Buzz"
+  else
+    puts x
+  end
+  x += 1
+end
```
mschae16, line 12: @estein1988 - This looks to be a great start! One part you may want to revisit is for the case where `x` is both a multiple of 3 and a multiple of 5?  I think in the current state it's jumping into the first if statement and implicitly returning afterwards.

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,28 @@ def initialize(protein, base, toppings)
...
+    "Please remove #{toppings}."
+  end
+
+  def change_protein(protein)
+    "I changed my mind I want #{protein}."
+  end
 end
 
 dinner = Burrito.new("Beans", "Rice", ["cheese", "salsa", "guacamole"])
+array = ["cheese", "salsa", "guacamole"]
```
mschae16, line 29: Have you tried passing the `array` variable to your declaration of the new Burrito instance?

day_7/10_little_monkeys.rb
estein1988: @mschae16

Being so new I copied over my code from FizzBuzz and made it fit to the 10 Little Monkeys rhyme (I knew it had worked so instead of re-inventing the wheel, I decided to use what I knew already). I should have taken a step back and examined if every block had a purpose. This morning I inserted a (#) on lines 41 and 42 and I agree they don't serve a purpose.

I will commit/push over the changes from your comments when everything is complete, thanks for taking a look at all of this!

day_6/exercises/burrito.rb
estein1988: @mschae16

I didn't know how to add/remove elements from the array since the array wasn't defined on its own (the ["cheese", "salsa", "guacamole"] array was nested within the Burrito class). This is why on line 29 I defined a different array, being array = [<3 original toppings elements>]. Then, on lines 34 & 36, elements can be added and removed to this "array". 

I inserted "toppings = " into line 28 and removed line 29. Is there a better way to add/remove elements to this array or is this the best way? 

```
dinner = Burrito.new("Beans", "Rice", toppings = ["cheese", "salsa", "guacamole"])
#####array = ["cheese", "salsa", "guacamole"]
p dinner.protein
p dinner.base
p dinner.toppings
p dinner.add_topping("lettuce")
p toppings.push("lettuce")
```

day_7/10_little_monkeys.rb
```diff
@@ -32,14 +32,14 @@
     He fell off and bumped his head,
     Mama called the doctor and the doctor said,
     \"Get those monkeys right to bed!\""
-  elsif n <= 9999999999
+  else n <= 9999999999
```
margoschaedel, line 35: `else` statements cannot include the conditional - the `n <= 9999999` part.

PR #346, opened 2020-06-10
John Kim

abcdefghijohn:
Here is my technical capstone. I wasn't able to finish the checker board activity because I ran out of time. My first week of mod_0 I was still working full time, I plan on working on it and finishing but thought I should turn in what I have at least.


francepack:
@abcdefghijohn 
Overall, good job with your capstone. A couple things I wanted to bring attention to:

- There are several files throughout missing work. Just a couple- I don't think its reason to hold you up... but remember to slow down at times and double check that you've hit everything asked for- that could save you some grief from Turing instructors!

- I'd encourage you to try every day 7 challenge! Starting with pseudocode, and writing your code a piece at a time is excellent practice for Mod1. A couple things I noticed:
1. In 10 little monkeys, you need some conditional logic to control when the different last line is delivered.
2. With any of these, you can practice receiving a user input like you do with your cipher. For example, you could ask for an input to determine how many times 10LM runs, or to set a range for fizzbuzz.

PR #347, opened 2020-06-16
Drew Williams

drewwilliams5280:
Here is my prework!


day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,16 @@
+numbers = [*1..100]
+numbers.map! { |num|
```
margoschaedel, line 2: What was the reason behind choosing the `.map!` method here?  That method is replacing the current `|num|` with the text output in the array, effectually mutating the originally `numbers` array which is usually something we want to avoid doing when programming.  Have you thought about trying to solve this with a `for` loop?

day_7/10_little_monkeys.rb
```diff
@@ -0,0 +1,12 @@
+monkeys = ["Ten", "Nine", "Eight", "Seven", "Six", "Five", "Four", "Three", "Two", "One"]
```
mschae16, line 1: Nice work architecting this solution!

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,27 @@ def initialize(protein, base, toppings)
...
 dinner = Burrito.new("Beans", "Rice", ["cheese", "salsa", "guacamole"])
 p dinner.protein
 p dinner.base
 p dinner.toppings
+dinner.add_topping("lettuce")
+p dinner.toppings
+dinner.remove_topping("salsa")
+p dinner.toppings
+dinner.change_protein("Chicken")
+p dinner.protein
```
mschae16, line 37: Nice work on this Burrito class!

PR #351, opened 2020-07-16
Jesse Mellinger

JesseMellinger:


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,76 @@
+i = 10
+
+until i == 0
```
damwhit, line 3: Each loops are definitely a bit more canonical in ruby. How could you utilize one here?

day_6/exercises/burrito.rb
```diff
@@ -1,19 +1,39 @@
...
     @toppings = toppings
   end
+
+  def add_topping(new_topping)
+    @toppings.push(new_topping)
+    @toppings
+  end
+
+  def remove_topping
+    @toppings.pop
```
damwhit, line 22: What topping are you removing here? How could you make it more dynamic?

PR #352, opened 2020-07-19
Taylor Phillips

taphill:
Here's my technical prework!

### DIFF
--- a/day_1/ex1.rb
+++ b/day_1/ex1.rb
@@ -1,0 +1,27 @@
+#
+# Exercise 1
+#
+puts "\nHello World!"
+puts "Hello Again"
+puts "I like typing this."
+puts "This is fun."
+puts "Yay! Printing."
+puts "I'd much rather you 'not'."
+puts 'I "said" do no touch this.'
+
+#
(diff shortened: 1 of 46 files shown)


francepack:
Nice work here! I wanted to leave a bit of feedback for consideration as you ramp up to start Turing:

- In general, I don't think you need a line for `attr_accessor` as well as one for `attr_reader`. I believe `attr_accessor` does everything `attr_reader` does, and then some, so you could just use `attr_accessor`. This allows not only reading the attribute, but also `writing` over it, as in modifying it. `attr_reader` does not allow writing over an attribute, so should be used in cases where you never want to modify an attribute.

- For speckled frogs, I think you could whittle down the lines of your solution by using a loop. That way, you can have some conditionals that just do something special for the last few verses, when words start being different. I don't think you should need to loop through every word to pick out those numbers... I'd suggest putting the numbers into your verses with interpolation.

- The next step on the checkerboard is to figure out how to construct the 2 different rows so the length of rows are the same as how many rows there are. A loop could help here, as well!

PR #354, opened 2020-07-21
Megan Campbell

Mcampbell26:


day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,18 @@
+maximum = 100
+tally = 0
+
+while tally <= maximum
```
damwhit, line 4: Good syntax and style here! .each is more canonical in ruby. How might you go about using .each here instead of a while loop?

day_6/exercises/burrito.rb
```diff
@@ -1,19 +1,44 @@
...
+
+  def add_topping(topping)
+    @toppings.push(topping)
+  end
+
+  def remove_topping(topping)
+    @toppings.delete(topping)
+  end
+
+  def change_protein(pro)
```
damwhit, line 26: Definitely prefer full words for arguments. I'd say `protein` would be more clear!

day_7/ceasar_cipher.rb
```diff
@@ -0,0 +1,45 @@
...
+#
+# a = Array("a" .. "z")
+#
+# a.rotate!(1)
+# a.zip(a.rotate!(1))
+#
+#
+# p a
+
+# this works to encrypt
```
damwhit, line 11: Nice job getting the encrypt portion working!

PR #355, opened 2020-07-21
Sean Steel

s-steel:
Here is my Mod 1 pre-work!

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -1,19 +1,45 @@
-# Add the following methods to this burrito class and 
+# Add the following methods to this burrito class and
 # call the methods below the class:
 # 1. add_topping
 # 2. remove_topping
 # 3. change_protein
 
 class Burrito
-  attr_reader :protein, :base, :toppings
+  attr_accessor :protein, :base, :toppings
   def initialize(protein, base, toppings)
     @protein  = protein
     @base     = base
     @toppings = toppings
+  end
+
+  def add_topping(toppings)
+    self.toppings = @toppings.push(toppings)
+    puts "Now your toppings include #{toppings}!"
+  end
+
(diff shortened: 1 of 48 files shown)


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,19 @@
+def speckled_frogs(num)
+
+  in_words = {0 => 'zero', 1 => 'one', 2 => 'two', 3 => 'three', 4 => 'four', 5 => 'five', 6 => 'six', 7 => 'seven', 8 => 'eight', 9 => 'nine', 10 => 'ten'}
+
+  until num == 0
```
damwhit, line 5: Arrays in ruby are 0 indexed, how could you accomplish the same thing with an array?

day_7/ceasar_cipher.rb
```diff
@@ -0,0 +1,26 @@
+def ceasar_cipher(string, shift)
+
+  alphabet = ('A'..'Z').to_a
+  encode = Hash[alphabet.zip(alphabet.rotate(-shift))]
+
+  new_array = string.upcase.chars.to_a
```
damwhit, line 6: What would be a better name for this array? What data is the array holding?

day_6/exercises/burrito.rb
```diff
@@ -1,19 +1,45 @@
...
   end
+
+  def add_topping(toppings)
+    self.toppings = @toppings.push(toppings)
+    puts "Now your toppings include #{toppings}!"
+  end
+
+  def remove_topping(remove_topping)
+    # you can also add "" to the toppings variable
+    self.toppings -= [@toppings.delete(remove_topping)]
```
damwhit, line 22: How could you simplify this statement? Do you need to reassign self.toppings? Same for the add_topping method above

PR #356, opened 2020-07-22
George Soderholm

GeorgieGirl24:
Here is my Pre-Work!!!

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -1,4 +1,4 @@
-# Add the following methods to this burrito class and 
+# Add the following methods to this burrito class and
 # call the methods below the class:
 # 1. add_topping
 # 2. remove_topping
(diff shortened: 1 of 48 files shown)


day_7/exercises/10_speckled_frogs.rb
```diff
@@ -0,0 +1,83 @@
+require "pry"
```
damwhit, line 1: Love that you're using pry!

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,45 @@ def initialize(protein, base, toppings)
...
+  def add_topping(*topping)
+    topping.each do |t|
+      self.toppings << t
+    end
+    self.toppings.sort
+    # self.toppings << topping
+
+  end
+
+  def remove_topping(*topping)
```
damwhit, line 24: The way this is setup, we're removing multiple toppings. What would be a better name in that case?

Alternatively, how could you simplify it to remove just one topping?

PR #357, opened 2020-07-22
Grayson Cunningham

cunninghamge:


day_6/exercises/burrito.rb
```diff
@@ -1,19 +1,38 @@
...
+    @toppings << t
+  end
+
+  def remove_topping
+    @toppings.pop
+  end
+
+  def change_protein(p)
+    @protein = p
+  end
```
ap2322, line 26: Nice work updating the attributes of the class and using ruby array methods.

day_7/checker_board.rb
```diff
@@ -0,0 +1,27 @@
...
+  s = size
+  if n.even?
+    while s > 0
+      if s.even?
+        print "X"
+      else
+        print " "
+      end
+      s = s - 1
+    end
```
ap2322, line 15: Pro-tip: loops inside of loops* can often be abstracted into it's own method that you can then call.

*though nested loops are inefficient computationally, they're often necessary for 2D problems like this one.

day_7/ceasar_cipher.rb
```diff
@@ -0,0 +1,36 @@
+class CeasarCipher
+  def encode(text, shift)
```
ap2322, line 2: Nice work thinking in terms of classes and methods within those classes. For mod 1, you'll get into a pattern of using an `initialize` method and then abstracting the other methods to use.

day_7/ceasar_cipher.rb
```diff
@@ -0,0 +1,36 @@
...
+      else
+        text_index << t
+      end
+    end
+
+    cipher = [*"A".."Z"]
+    shift.times do |x|
+      x = cipher.shift
+      cipher << x
+    end
```
ap2322, line 19: I'm not sure this is accomplishing a shifted alphabet in the way you think. Check out ruby's `rotate` method to see if it's shifting your alphabet more consistently.

As a general rule, you'll want to stay away from naming variables the same as ruby methods `shift` in this case.

ap2322 (commented):
@cunninghamge overall, excellent work on this! Great job pushing yourself on both of the optional day 7 exercises. That type of work ethic always leads to success at Turing!

PR #360, opened 2020-07-22
JAKE HEFT

jakeheft:


day_7/exercises/fizzbuzz.rb
```diff
@@ -0,0 +1,58 @@
+# Once finished, another student and I compared code and I realized that defining
+#   mult_3 & mult_5 was unnecessary math that could have been done inside fizz_buzz
+#   method at each if statement. I left it as is because that was my original work;
+#   not sure to what degree borrowing code from someone else is allowed for assignments
+def mult_3(num)
```
damwhit, line 5: I like the method abstraction here and below. Way to DRY it up!

damwhit: Remember that a comparison statement returns true or false. How would that simplify this method and the one below?

PR #362, opened 2020-07-22
Curtis Bartell

c-bartell:
Here is my technical prework.


day_7/projects/fizzbuzz.rb
```diff
@@ -0,0 +1,57 @@
...
+  # end loop
+# end method
+# Call fizz_buzz and pass 100 to it
+
+def fizz_buzz(range_min, range_max)
+
+  current_number = range_min
+  output = ''
+
+  while current_number <= range_max
```
damwhit, line 31: How could you use a .each loop here instead? That's typically a more canonical way to do things in ruby

PR #363, opened 2020-07-22
Cydnee Owens

cowens87:
Here is Mod 1 Pre-work


francepack:
Nice work here. Just a bit of feedback as you ramp up towards the start of Turing:
- For your cipher, you could get the amount to displace each letter by a gets.chomp as well.

- Nice strategy for fizzbuzz, just remember the indentation in your code blocks help this be read well! Also, you could use gets.chomp to get some user input to make the amount of times this runs variable.

- I'd encourage trying some extensions for speckled frogs! You could try to use a hash to hold some values that help you translate your integers into spelled out numbers like 'one', 'two', etc.

PR #364, opened 2020-07-22
Caryn Keffer

carynkeffer:


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,31 @@
+class Frog
+
+  def initialize
```
damwhit, line 3: If a method is empty. No need for it!

day_6/exercises/classes_and_objects_worksheet.rb
```diff
@@ -0,0 +1,388 @@
+# this file acts as a personal worksheet for myself to help me go through the
+# lesson from launchschool: 'Classes and Objects - Part 1'
+
+# ***** States and Behaviors
+
+# - Remember that we use classes to create objects.
```
damwhit, line 6: Good notes here

PR #366, opened 2020-07-24
Connor Ferguson

cpfergus1:


day_7/fizzbuzz/fizzbuzz.rb
```diff
@@ -0,0 +1,18 @@
+puts "What number would you like me to count to?"
+count_to = gets.chomp.to_i
+i=1
+print_message = ""
+puts "Here... We... Go!"
+while i <= count_to
```
damwhit, line 6: Prefer .each in ruby to while loops

day_7/speckled_frogs/10_speckled_frogs.rb
```diff
@@ -0,0 +1,53 @@
+#credit to:
+#user "NoNonsense"
+#October 6, 2014
+#https://stackoverflow.com/questions/19445003/using-ruby-convert-numbers-to-words
+#for number conversion
```
damwhit, line 5: @cpfergus1 remember that it's ok to look at other code for inspiration. But we don't want you wholesale copy pasting code you find from the internet.

day_7/ceasar_cipher/ceasar_cipher.rb
```diff
@@ -0,0 +1,49 @@
+# Thanks to matugm @ https://gist.github.com/matugm/db363c7131e6af27716c
+# for this method of running a ceasar cipher.
+
```
damwhit, line 3: Same thing here... Inspiration is ok, but we don't want you copy/pasting. The end product is often less important than struggling through a problem to deeply learn what it takes to solve it

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,30 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(new_topping)
+    @toppings = @toppings + [new_topping]
+  end
+
+  def remove_topping(x_topping)
```
damwhit, line 19: Good param naming here

PR #367, opened 2020-07-27
James Fox-Collis

jlfoxcollis:
Here is my prework!  Looking forward to receiving any feedback and continuing to learn Ruby!

Thank you!

James

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -1,4 +1,4 @@
-# Add the following methods to this burrito class and 
+# Add the following methods to this burrito class and
 # call the methods below the class:
 # 1. add_topping
 # 2. remove_topping
@@ -11,9 +11,26 @@ class Burrito
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping
+    toppings << "Sour Cream"
+  end
+
+  def remove_topping
+    toppings.pop(2)
+  end
+
+  def change_protein
+     protein << "Fish and beans"
+  end
+
 end
 
+
+
 dinner = Burrito.new("Beans", "Rice", ["cheese", "salsa", "guacamole"])
+dinner.add_topping
 p dinner.protein
 p dinner.base
 p dinner.toppings
+p dinner.remove_topping
--- a/day_7/10_speckled_frogs.rb
+++ b/day_7/10_speckled_frogs.rb
@@ -1,0 +1,65 @@
+# frog_number = {
+#   1 => "one",
+#   2 => "two",
+#   3 => "three",
+#   4 => "four",
+#   5 => "five",
+#   6 => "six",
+#   7 => "seven",
+#   8 => "eight",
+#   9 => "nine",
+#   10 => "10"
+# }
+
+frog_number = {
+  10 => "ten",
+  9 => "nine",
+  8 => "eight",
+  7 => "seven",
+  6 => "six",
+  5 => "five",
+  4 => "four",
+  3 => "three",
+  2 => "two",
+  1 => "one"
+}
+
+def plural_frog(key)
+  (key == 1) ? "frog" : "frogs"
+end
+
+def plural_frogs(number_2)
+  (number_2 == 1) ? "frog" : "frogs"
+end
+
+def were_was(number_2)
+  (number_2 == 1) ? "was" : "were"
+end
+
+
+# whole_number = (1..10).to_a.reverse
+
+
+frog_number = frog_number.map do |key, value|
+
+  number_1 = value
+  number_2 = key - 1
+
+
+        puts "#{number_1} speckled #{plural_frog(number_1)} sat on a log eating some most delicious bugs.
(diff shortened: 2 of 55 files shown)


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,65 @@
...
+  6 => "six",
+  5 => "five",
+  4 => "four",
+  3 => "three",
+  2 => "two",
+  1 => "one"
+}
+
+def plural_frog(key)
+  (key == 1) ? "frog" : "frogs"
```
ap2322, line 28: Gettin' fancy with the ternary operator

day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,65 @@
...
+
+def were_was(number_2)
+  (number_2 == 1) ? "was" : "were"
+end
+
+
+# whole_number = (1..10).to_a.reverse
+
+
+frog_number = frog_number.map do |key, value|
```
ap2322, line 43: Good instinct to use iteration and nice work using the hash block variable syntax with `|key, value|`. 

You don't gain anything by reassigning `frog_number` on this line or using `.map` instead of `.each`. Take a look at the difference between those two enumerators to see why you'd get the same output with `.each` in this implementation.

day_7/caesar.rb
```diff
@@ -0,0 +1,34 @@
...
+# moving = encode.map {|x| x + 3}
+# moving.map {|x| x.chr }.join
+# end
+#
+# puts to_code("ouch")
+
+def caesar_cipher(k)
+  alphabet = Array('a'..'z')
+  encrypter = Hash[alphabet.zip(alphabet.rotate(3))]
+  k.split('').map { |x| encrypter.fetch(x, " ") }
```
ap2322, line 15: Very nice work creating a hash with the enumerator .zip and using `alphabet.rotate` then using .map on the next line to reach into the hash and create an encrypted array. There's a lot happening on these two lines!

day_7/caesar.rb
```diff
@@ -0,0 +1,34 @@
...
+end
+
+
+puts "Feed me your Password!(Or anything.)"
+password = gets.chomp
+
+puts "YUMMY!"
+p caesar_cipher("#{password}").join
```
ap2322, line 23: If the goal of the encrypter is to output a string, the `.join` step would conventionally be the last step in the `caesar_cipher` method.

day_7/caesar.rb
```diff
@@ -0,0 +1,34 @@
...
+
+
+def decoder(k)
+  alphabet = Array('a'..'z')
+  decrypter = Hash[alphabet.zip(alphabet.rotate(-3))]
```
ap2322, line 28: Great work thinking through the next step of decrypting. How might you refactor encrypt/decrypt methods that use `rotate(-3)` to not always tied to shifting 3 places?

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+
+  if x % 5 == 0
+    puts "buzz"
+  end
+
+  if x % 3 == 0 && x % 5 == 0
+    puts "fizzbuzz"
+  end
```
ap2322, line 15: It's an antipattern here to disconnect your conditional statements when considering multiple scenarios for x. Also, what would happen when x == 15?

day_7/frogs.rb
```diff
@@ -0,0 +1,81 @@
...
+          number_2 = number_2 % key
+        end
+      end
+    end
+    number_word_one = original_words.join(' ')
+    number_word_two = subtraction_words.empty? ? 'no more' : subtraction_words.join(' ')
+    puts "number: #{number}"
+    puts "#{number_word_one} speckled #{pluralize_frog(number)} sat on a log eating some most delicious bugs. "
+    puts "One jumped in the pool where its nice and cool, "
+    puts "then there #{were_was(number - 1)} #{number_word_two} speckled #{pluralize_frog(number - 1)}."
```
ap2322, line 77: There's a lot going on here, good job finding a way to handle changing number, converting numbers to words, and building the string to puts out. As you continue programming, if you find yourself with discrete iterations inside of a function - it could be a good time to pull out a helper method.

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,26 @@ def initialize(protein, base, toppings)
...
+  def add_topping
+    toppings << "Sour Cream"
+  end
+
+  def remove_topping
+    toppings.pop(2)
+  end
+
+  def change_protein
+     protein << "Fish and beans"
```
ap2322, line 24: What does running `puts dinner.change_protein` result in when `protein == "Beans"`?

ap2322 (commented):
@jlfoxcollis good work overall on this. Really great work exploring the built-in ruby methods and using those tools to implement solutions. Keep practicing and pushing your skills!

PR #369, opened 2020-07-28
Aidan Murray

TeknoServal:


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,24 @@
+require 'humanize'
```
mschae16, line 1: Nice work implementing a `while` loop for this exercise! One easy way to make this a little less repetitive is to store some of the rhyme copy in variables, and then reference the variables within the loop. It also makes things a little easier to read.  And while it's impressive that you have a solid enough grasp on ruby to be comfortable pulling in external gems/libs (`humanize`), I would highly recommend keeping things simple for now and trying to achieve the desired result with your own code alone.

PR #372, opened 2020-07-29
Shaunda Cunningham

smcunning:


francepack:
Nice work here! Just a bit of feedback as you keep working, ramping up to the start of Turing:
- In your Person class, I wouldn't use a method to add an attribute like this. Right now, `interests` doesn't exist until `add_interests` is run. If `introduction` was run before `add_interests` was, it wouldn't know where to get the interests it tries to list.

- Nice job on the day 7 work! I'd keep playing with these- there are many solutions, and the checker board is pretty tough, so I'd recommend trying it! Nice job finding and using some approachable ruby methods `downto` and `upto`.

PR #376, opened 2020-07-31
Hanna Davis

Oxalisviolacea:
Module 1 Prework. Thanks!


francepack:
Nice work on this! Here is some feedback to consider as you keep working, ramping up for the start of Turing:
- I like the concepts of your person class, but looking at the method names, I have a harder time understanding why they do the things they do. Particularly method `know` that reassigns attribute `code` to `false`.  Method `learn` makes a ton of sense, but `know` also sounds like it'd do something similar. Maybe something like `forget`? 

- Nice solutions for day 7! Some of those are pretty advanced, so if some research went into using some of the methods throughout there, I'd be sure to double check understanding what is happening line-by-line, particularly for your ceasar cipher- lots going on line by line there!

- Awesome work with your pseudocode!

PR #377, opened 2020-08-26
Finished Prework (I think)

b-enji-cmd:


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,8 @@
+puts "How many frogs would you like to start with?"
+frog_num = gets.chomp.to_i
+
+for i in (frog_num).downto(1)
+
+  puts "#{i} speckled frogs sat on a log\neating some most delicious bugs.\nOne jumped in the pool where its nice and cool,\nthen there were #{i-1} speckled frogs.\n\n "
```
damwhit, line 6: I believe the phrase changes as the numbers go down. How would you reflect that in your code

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,32 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(new_topping)
+    toppings.push(new_topping)
```
damwhit, line 16: why are you able to leverage this without the @ symbol prefixing toppings?

damwhit (commented):
@b-enji-cmd solid work on this

day_6/exercises/burrito.rb
b-enji-cmd: I believe it has something to do with how I have the attr_reader, had I used attr_accessor I wouldn't need to use the @ on any of them. Is this correct?

day_7/10_speckled_frogs.rb
b-enji-cmd: I would probably have a conditional statement when the current iteration is equal to 1. Is that what you were referring to?

damwhit: Yeah, I think that would accomplish it

day_6/exercises/burrito.rb
damwhit: correct!

PR #378, opened 2020-09-03
Max Ribbans

ribbansmax:
Here is my prework!


francepack:
@ribbansmax 

- In your classes, I see you use `self.` quite a bit... This works, but I more often see `@` used in those situations, and it is good to keep consistency throughout.

- Nice OOP solution to speckled frogs! I appreciate the care you take to set yourself up with the information you'll need to reference in speckled frogs and the cipher.

- For fizzbuzz, I am not sure you will need a temp counter. What you are going for there is pretty clever, but I think sticking to a conditional statement not based on a temp variable would be more clear. I do like how you are thinking to construct `FizzBuzz` out of 2 different conditions applying, however.

PR #379, opened 2020-09-05
Joe Jiang

ninesky00:
Here is my Prework.  
Link to Gear up: https://gist.github.com/ninesky00/80346e3b7fa81bcb2044ea98cf310fff  
Link to journal: https://gist.github.com/ninesky00/8b31c5b92a2ef82a091cf0d5fd46367b  
Link to prework repo: https://github.com/ninesky00/backend_mod_1_prework


francepack:
@ninesky00 

- Great work utilizing a ruby gem, humanize, to assist with speckled frogs. It is also a great mental exercise to consider how you might accomplish this without humanize, or think about what the inner workings of humanize might be. I looked forward in your work, and now see you've done this in the num_to_words file- excellent work, and instinct to be inquisitive!

- Nice work setting up each challenge as a simple ruby program, gathering user input when required.

PR #381, opened 2020-09-08
Jesus Quezada-Guillen

jequeza:
Here is my pre-work!


francepack:
@jequeza 

- I like the start you have for the checkerboard. Next step- how can you create those 2 different types of rows with variable length? You could construct the 2 different rows you use based on the user input, then keep the same technique you use below.

- For fizzbuzz, the order you have your conditionals can make an impact. Right now, I'd expect any numbers that potentially meet your second condition to first get swept up in the first condition. I'd bring your or statement to the top, here, to make sure that any this one applies to enter that conditional block.

PR #382, opened 2020-09-08
Aiden Mendez

aidenmendez:
Here is my pre-work!

### DIFF
--- a/day_7/10_speckled_frogs_extension_2.rb
+++ b/day_7/10_speckled_frogs_extension_2.rb
@@ -1,0 +1,27 @@
+puts "Enter the number of frogs:"
+print "> "
+count = gets.chomp.to_i
+next_count = count - 1
+
+while count > 2
+   puts "#{count} speckled frogs sat on a log eating some most delicious bugs."
+   puts "One jumped in the pool where its nice and cool"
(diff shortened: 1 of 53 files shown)


day_7/fizzbuzz_bonus.rb
```diff
@@ -0,0 +1,31 @@
...
+
+if top_range <= bottom_range
+    puts "Invalid range! Exiting program."
+    exit
+end
+
+
+
+while bottom_range <= top_range
+    if bottom_range % 3 == 0 && bottom_range % 5 == 0
```
damwhit, line 17: prefer 2 spaces to 4 for indents

day_7/10_speckled_frogs_extension_2.rb
```diff
@@ -0,0 +1,27 @@
+puts "Enter the number of frogs:"
+print "> "
+count = gets.chomp.to_i
+next_count = count - 1
+
+while count > 2
+   puts "#{count} speckled frogs sat on a log eating some most delicious bugs."
```
damwhit, line 7: how could you refactor so that this statement doesn't need to be duplicated below?

damwhit (commented):
@aidenmendez solid work here!

PR #385, opened 2020-09-09
Joseph Budina

josephbudina:
This is my prework!


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,38 @@
+def frog(num)
```
damwhit, line 1: nice abstraction of methods here!

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,13 @@
+num = (1..720)
+
+num.each do |i|
```
damwhit, line 3: love this. Super clean!

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,24 @@ def initialize(protein, base, toppings)
...
+  def add_topping
+    @toppings.push("lettuce")
+  end
+
+  def remove_topping
+    @toppings.delete("cheese")
+  end
+
+  def change_protein
+    protein.replace("steak")
```
damwhit, line 24: Why are you able to use protein without the @symbol here?

damwhit (commented):
@josephbudina solid work on this! You are officially `turing ready`!

day_6/exercises/burrito.rb
josephbudina: I want to say that it's because that value of protein is just being read and not actually changed. I ran some tests with pry trying to read the values, but I keep getting an error message about burrito not being a method rather than it giving me the values.
Screen Shot 2020-10-03 at 5 02 21 PM

josephbudina: I did find that I was able to remove the @ symbol from all the methods and receive the same result as well, which I found interesting. I'm continuing to look into it, but at least wanted to respond to you before class on monday.

PR #391, opened 2020-09-11
Jordan Beck

jordanfbeck0528:
Carreer Journal - 
https://gist.github.com/jordanfbeck0528/381e1a89ec48d1422e90fe2891f85482#file-career_journal-md
Gear up -
https://gist.github.com/jordanfbeck0528/5862a73a34ab1f19580cab251a296708
Technical -
https://github.com/jordanfbeck0528/backend_mod_1_prework


francepack:
@jordanfbeck0528 
- With your burrito class, you've made methods `add_toppings` and `remove_toppings`, but I see when you call those methods, they just take a sopping string as a parameter. I'd name those methods singularly, unless the parameter you give them is an array with multiple toppings. Nice job with your setter methods in your person class!

- With your speckled frogs solution, it is pesky, but there is a word that should be substituted when `i == 2`. - the last line should have `was one speckled frog` instead of `were one speckled frogs`. As a dev, you have a decision on whether to have the whole line by line poem in another condition block, or perhaps break out that last line to reconstruct for the 3 different conditions. Always try to avoid re-typing something that will be the same for all conditions, such as the 2nd line of the poem.

- Nice work with fizzbuzz. Here, you may think about how you can leverage the order of your conditions to cut out a little bit of typing on your end. If you have the condition you currently have last first instead, you will only need to check one condition for anything that doesn't enter the `x % 3 == 0 && x % 5 == 0` condition. Basically, if this condition is first, you do not need check for anything being false in the next 2 conditionals.

jordanfbeck0528:
- Changed all uses of `toppings` to `topping`

- Modified answer to use proper grammar for Poem

- Refactored fizzbuzz

PR #392, opened 2020-09-12
Samuel Yeo

SK-Sam:
Mod 1 Prework Complete!


francepack:
@SK-Sam 

- Nice work choosing a strategy for your cipher, and putting in the research to make it happen! I really like the decisive approach- I find it helpful to choose an approach and go with it. If it doesn't end up working quite right, or best meeting your needs, at least you have just now found that out.

- Also well done formatting the solutions to your challenges as interactive ruby programs, getting an input from a user, etc. You could do this with your cipher as well!

PR #394, opened 2020-09-15
Vanessa Alves de Arruda

nessaarruda:
Here is my prework!


francepack:
@nessaarruda 

- The methods for your human class all grab an attribute, and return a string based on the value in that attribute. When I see a method like `addName`, I don't intuitively think 'This is a method that will make a string based on the name attribute', I think 'this is a method that will give my human a name.' I know we have discussed 'setter' vs 'getter' methods elsewhere, so I am sure this is getting clearer now, but one more reminder to focus on describing specifically what you are doing with your method names, for example `printName` versus `changeName`- one of those is a getter method, one of those is a setter method that will change the value of an attribute.

- You might think about leveraging an object, or an array, to help with substitution for those integers. You can perhaps get different elements of an array with spelled out number strings by index based on the increment of a loop.

nessaarruda:
@francepack Hey, thank you for the feedback. I did change the methods to getters. For the second piece of feedback. I think know how to make it work, it would be something like:
``` my_numbers = ["one", "two", "three"]
new_array = []

my_numbers.each do |num|
  new_array << num 
  print new_array
end  ```

I'm just not sure which exercise you are referring to.

PR #395, opened 2020-09-16
Dominic Padula 

domo2192:


francepack:
@domo2192 

- With your burrito class, you've made methods add_toppings and remove_toppings, but I see when you call those methods, they just take a sopping string as a parameter. I'd name those methods singularly, unless the parameter you give them is an array with multiple toppings. Nice job with your setter methods in your person class!

- The methods for your human class all grab an attribute, and return a string based on the value in that attribute. When I see a method like addName, I don't intuitively think 'This is a method that will make a string based on the name attribute', I think 'this is a method that will give my human a name.' I know we have discussed 'setter' vs 'getter' methods elsewhere, so I am sure this is getting clearer now, but one more reminder to focus on describing specifically what you are doing with your method names, for example printName versus changeName- one of those is a getter method, one of those is a setter method that will change the value of an attribute.

- I do not see your solution attempt for Speckled frogs. I saw a start of a number converter to help you get spelled out versions of numbers, but you can leave anything like that for after you have a basic working solution with numbers. I highly encourage you to keep at some of the day 7 challenges! This will be great mental exercise leading up to your Turing start date.

PR #396, opened 2020-09-17
Yesi Meza

Yesi-MC:
Here is my Mod 1 prework

### DIFF
--- a/day_1/README.md
+++ b/day_1/README.md
@@ -48,13 +48,13 @@ This will open the day_1 directory in Atom. You should be able to see the direct
 
 1. Check off the items below as you complete the steps you just read for each lesson. ***Remember to create a file containing your work for each lesson!***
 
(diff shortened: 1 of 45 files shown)


francepack:
@Yesi-MC 

- With your person class, I may expect a method like `talk` to report back a message of some sort, rather than just change an attribute. You could perhaps have an attribute called `greeting`, and a method `talk` logs your greeting to the console. You could have another method called `change_greeting` that directly modifies that `greeting` attribute. Likewise, right now, what your `get_nickname` method does is change the name attribute based on a user input, so to me this sound more like a method `change_name`. Basically, I really like how you are making 'setter' methods that modify attributes, but I think more accurate naming could help bring across what specifically they do.

- Nice work seeing the different conditions to hit in speckled frogs! In your second 2 condition, I would just use a `==` operator to be more clear that they will only be used at that single frog number. You could also think about trying to refactor, removing any pice that is reused multiple times.

PR #398, opened 2020-10-25
Kyle Schulz

kylejschulz:


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,25 @@
...
+    some most delicious bugs.
+    One jumped in the pool where its nice and cool,
+    then there were #{(frog - 1).humanize} speckled frog."
+  else
+    puts "#{frog.humanize} speckled frogs sat on a log
+    some most delicious bugs.
+    One jumped in the pool where its nice and cool,
+    then there were #{(frog - 1).humanize} speckled frogs."
+  end
+end
```
francepack, line 25: Nice work here- one thing you can think about is breaking repeated lines into a variable, so you just have to type the end of the 1st line, and the 2nd/3rd lines once. Whenever you see repeated words, or logic, that can be a sign to break something out into a variable or method!

day_7/ceasar_cipher.md
```diff
@@ -14,3 +14,8 @@ cipher = CeasarCipher.new
 cipher.encode("Hello World", 5)
 => "CZGGJ RJMGY"
 ```
+prompt user for code and how many steps to right or left
+
+turns the string into an array of numbers
+make alphabet array
```
francepack, line 20: We were looking to see one of either the cipher or the checkerboard coded out, as per the instructions. We aren't expecting perfection here, but these challenges are great ways to practice breaking problems apart, like you've started here, and implementing some research into how to achieve those steps. I'm seeing enough from your class work, speckled frogs, and fizzbuzz to know you're on the right track as far as material we want students to enter Turing with, but I highly encourage trying out BOTH the cipher and the checkerboard in the weeks leading up to your start date as excellent practice. Feel free to drop me a line via slack if you do add to your work here, and would like feedback on that.

francepack: I looked at the repo again, and I see coding these out is optional (this is a change). Regardless, I encourage you to give them a shot! Just wanted to reply here that I was mistaken that coding them out was asked of you.

kylejschulz: Mason, thanks for the feedback. I submitted my caesar cipher and would love some feedback. 

a few questions i had:
-should i make it able to accept phrases as well?
-is there a way to refactor the code so it's more readable, neat, and has better separation of functionality? I still get a little confused as to where i need to define things in the class.
-are there any important tools or methods i'm missing that might make this easier? 

i'm starting to work on the checkerboard now.

day_7/ceasar_cipher_attempt.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def initialize
+  end
+
+  def encode(word, shift_num)
+    #make aplphabet into array
+    alphabet = 'abcdefghijklmnopqrstuvwxyz'.split('')
+    #make encoded alphabet array
+    encoded_alphabet = alphabet.unshift(alphabet.last(shift_num + 1)).flatten(1)
+
```
francepack, line 11: Interesting strategy here! So what I am gathering is that you are making a second alphabet array with all of the letters shifted, then you grab the letters of the same index... you may be able to lean on just your original alphabet array here- you could do the 'shift' on your index, but you'll need to account for when that shift takes things beyond the number of elements in your array (like when a z gets shifted several letters).

day_7/ceasar_cipher_attempt.rb
```diff
@@ -0,0 +1,41 @@
...
+cipher.encode('abcde', 2)
+
+  # alphabet.each do |letter|
+  #   alphabet_index.push(alphabet.index(letter))
+  # end
+  #
+  # cipher_word.push(alphabet[(alphabet.index(letter) + shift_num)])
+
+  # encoded_array.push(encoded_alphabet[num])
```
francepack, line 41: Right now, when I run this, the shift goes backwards, so an 'a' shifted 2 becomes a 'y'. Traditionally, they shift forward, but totally developers decision! An interesting extension could be make a ruby program that grabs some user input to define the shift, as well as whether it shifts forwards or backwards. This is above and beyond here, but an idea of something to play around with as you continue to experiment.

day_7/ceasar_cipher_attempt.rb
```diff
@@ -0,0 +1,41 @@
+class CeasarCipher
+
+  def initialize
+  end
+
```
francepack, line 5: I like an OOP approach to this problem! Not having any attributes makes this a bit less OOPy.... its not wrong, but I think you could experiment with passing info into your instance creation to give your class a few attributes to be held there. But, ultimately, up to you!

PR #399, opened 2020-10-26
Olivia Charles

ochar721:
Attached is my prework.


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,27 @@
...
+
+def speckled_frogs
+  @range.each do |frog|
+    if frog >= 3
+      puts "#{frog} speckled frogs sat on a log eating some most delicious bugs. One jumped in the pool where its nice and cool then there were #{frog-1} speckled frogs."
+    elsif frog >=2
+      puts "#{frog} speckled frogs sat on a log eating some most delicious bugs. One jumped in the pool where its nice and cool then there were #{frog-1} speckled frogs."
+    elsif frog >= 1
+      puts "#{frog} speckled frogs sat on a log eating some most delicious bugs. One jumped in the pool where its nice and cool then there were #{frog-1} speckled frogs."
+    end
```
ap2322, line 25: Good work using iteration. It looks like your puts statements are all the same, though they are set up in conditionals that would allow for different wordings.

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,27 @@
...
+    puts number
+  end
+end
+
+#bonus
+puts "Enter any number"
+num = gets.chomp.to_i
+
+num.times do |number|
+  if number % 3 == 0 && number % 5 == 0
```
ap2322, line 18: Nice work on the bonus to iterate on a given number!

day_7/checker_board.rb
```diff
@@ -0,0 +1,14 @@
...
+
+number.times do |i|
+ if number > 1
+   puts ("X"*number)[i,number] + "\n"
+ elsif number == 1
+   puts " X X X "
+ else
+   puts "Try Again"
+ end
+end
```
ap2322, line 14: This is a good start to building the checkerboard. Remember you need to keep track of building rows all of the same length. Try drawing out a checkerboard one row at a time on a piece of paper and write down the steps you needed for the manual process. Then use that more detailed pseudocode to build your code algorithm.

day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,21 @@ def initialize(protein, base, toppings)
...
   end
+
+  def add_topping
+  end
+
+  def remove_topping
+  end
+
+  def change_protein
+  end
```
ap2322: Complete this exercise.

day_7/ceaser_cipher.rb
```diff
@@ -0,0 +1,17 @@
...
+ALPHABET_SIZE.times do |shift|
+  shiftArray << charLine.map do |c|
+    ((c + shift) < 123 ? (c + shift) : (c + shift)- 26).chr
+  end.join
+end
+
+  shiftArray
+end
+
+puts caesar_cipher("Grinch")
```
ap2322: This is clearly a copy-paste of [someone else's code](https://gist.github.com/matugm/db363c7131e6af27716c). Delete and replace with your own pseudocode in comments that explain each step so you can code the solution and fully understand it.

ap2322 (commented):
@ochar721 Looking through, it looks as though some exercises are missing. I've left more detailed feedback with request for updates below.

day_7/fizzbuzz.rb
ochar721: thanks. took me a long time to figure out

day_7/ceaser_cipher.rb
ochar721: sorry i was trying to understand what the code was doing and work it out line by line and must have had to stop. i will work on these today.

day_6/exercises/burrito.rb
```diff
@@ -12,20 +12,27 @@ def initialize(protein, base, toppings)
...
-  end
+def remove_topping(guacamole)
+  self.toppings.delete(guacamole)
+end
 
-  def change_protein
-  end
+def change_protein(protein)
+  @protein = protein
+ end
```
ap2322: Good work on the `change_protein` method that directly updates an `@instance_variable` using an argument. Because `protein` is also an `attr_reader` method that calls the `@protein` attribute, it might be worth it to alter the argument name slightly so you can tell the difference between the two.

Question regarding your toppings: Will the topping updated always be `corn`? Or always delete `guacamole`? What might be a better name for your argument for these methods?

Note on `self`: Rarely will you ever need to use `self` - especially in mod 1. What might be a different way to access the toppings attribute directly?

ap2322:
@ochar721 good work making these updates, you're officially `Turing Ready`. I'd encourage you to continue practicing pseudocoding problems before writing the code itself. 

Also, keep working on the [ruby-exercises](https://github.com/turingschool/ruby-exercises ) - especially mythical creatures in preparation for mod 1. Let me know if you have any questions.

PR #400, opened 2020-10-26
Adam Bowers

Pragmaticpraxis37:
Here is my prework!


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,49 @@
...
+      puts "then there were no more speckled frogs!"
+    elsif frog == 2
+      puts "then there was #{frog - 1} speckled frog!"
+    else
+      puts "then there were #{frog - 1} speckled frogs!"
+    end
+  end
+end
+
+speckled_frogs(numbers)
```
mschae16, line 23: Nice work on this exercise! Glad to see you did the extension as well. One thing I noticed here is that when testing your code, the `10_speckled_frogs` rhyme printed out from 1 - 10 rather than 10 counting down to 1 as requested in the instructions.  Can you think of a way to loop through the array of numbers in the reverse?

mschae16 (commented):
@Pragmaticpraxis37 - Overall, excellent work on this! I'm glad to see you pushed yourself and tried both of the optional day 7 exercises. That type of work ethic always leads to success at Turing!

day_7/10_speckled_frogs.rb
Pragmaticpraxis37: Thanks for feedback and the encouragement.  To loop through the array going from 10 down to 1, I would use the .reverse method when the variable is assigned.  I tried that in the code and it works.

PR #403, opened 2020-10-27
Jacob Volpe

javolpe:
Here is all of my pre-work!!


day_6/exercises/practice.rb
```diff
@@ -0,0 +1,30 @@
...
+    @name = n
+    @height = h
+    @weight = w
+  end
+
+  def speak
+    "#{name} says Arf!"
+  end
+
+  def change_info(n, h, w)
```
francepack, line 14: It can seem tempting to make multi purpose methods like this, but I'd try to go for single purpose methods. Nothing is wrong here, its just that everytime you want to change one of those things, you need to input all 3.

day_6/exercises/student.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def favorite_number
+    7
+  end
+end
+
+frank = Student.new
+frank.first_name = "Frank"
+frank.introduction("Katrina")
```
francepack, line 15: Interesting note- since you do not have an initialize method, you are actually first defining the attribute `first_name` on the instance of your student when you call it.

day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,38 @@
...
+    puts "\n" + rhyme2.reverse.sub('sgorf', 'gorf').reverse.gsub('were', 'was')
+    x -= 1
+
+  else puts
+    rhyme2 = rhyme.sub("#{x}", numbers_capital[x]).sub("#{y}", numbers[x-1])
+    puts rhyme2
+    x -= 1
+
+end
+end
```
francepack, line 38: Nice work here! A next step would be to leverage gets.chomp to get some input from users, then define your x and y based on that input.

day_7/high_level.md
```diff
@@ -0,0 +1,23 @@
...
+
+First thing I did was create the method to ask the user for input.  If the user inputs an odd number the method reruns itself until an even number is inputed.  If a string is entered the program ends. I saved the input from that method in a global variable then set two global variables to equal "X " and " X" to make my rows and columns filler.  
+
+Then I created another method using the saved user input as a parameter to print out the board.  The inputed number is divided by 2 at three points in the last method because my method defaults to printing two rows every iteration with two columns also filled each iteration.  I feel this if fine because checkerboards are always an even number of rows/columns.  I probably could reconfigure some things to make odd numbers of rows/columns acceptable.  Mainly saving the variables as `"x"`  and `" "` and then reconfiguring how it reads the block in the while loop and the while loop itself.  For now I'm going to leave it for the above reason.
+
+####Update
+I went back and changed the code to allow for odd numbers to be accepted.  Personally this goes against my beliefs on how a checkerboard is defined but life is about growth.
```
francepack, line 23: Excellent mindset! I know you wrote that kind of in jest, but it is true that we are learning to bend our programs to our will, so the extent to which they replicate reality is up to us!

day_7/high_level.md
```diff
@@ -0,0 +1,23 @@
...
+
+Ok at this point we have all the string, broken down into individual characters, index position found, X number of index positions added (or subtracted), found the index position of the new integer, and converted back to the value associated with that index position.  To return it back to one beautiful non-sensical string the `.join` method will have to be used **AFTER** all that has been completed and we should get an output of an encrypted message.
+
+
```
francepack, line 11: Really great stuff in your pseudocode- I hope you found this helped keep thoughts organized!

PR #404, opened 2020-10-27
Kris Litman

krislitman:
Here's my pre-work, thank you!

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -1,4 +1,4 @@
-# Add the following methods to this burrito class and 
+# Add the following methods to this burrito class and
 # call the methods below the class:
 # 1. add_topping
 # 2. remove_topping
@@ -11,9 +11,25 @@ class Burrito
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(new)
+     @toppings + new
+  end
+
+  def change_protein(newprotein)
+    @protein = newprotein
+  end
+
+  def remove_topping(r)
+    @toppings.delete(r)
+  end
+
 end
 
 dinner = Burrito.new("Beans", "Rice", ["cheese", "salsa", "guacamole"])
-p dinner.protein
+# p dinner.protein
+dinner.change_protein("Chicken")
+dinner.remove_topping('cheese')
 p dinner.base
-p dinner.toppings
+p dinner.protein
+p dinner.add_topping(["corn salsa"])
--- a/day_7/10_speckled_frogs.rb
+++ b/day_7/10_speckled_frogs.rb
@@ -1,0 +1,47 @@
+# 3 speckled frogs sat on a log
+# eating some most delicious bugs.
+# One jumped in the pool where its nice and cool,
+# then there were 2 speckled frogs.
+
+# 2 speckled frogs sat on a log
+# eating some most delicious bugs.
+# One jumped in the pool where its nice and cool,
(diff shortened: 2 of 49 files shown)


day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,25 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(new)
+     @toppings + new
```
francepack, line 16: When I see the `+` operator, I think of string concatenation, or math with integers. But toppings is an array. You may try some research into methods that work with arrays, much like the hmwk from tasks 1!

day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,50 @@
...
+  remaining_frogs = frog.to_i - 1
+  was_when = "were"
+
+  until frog == 2 do
+    p "#{frog} speckled frogs sat on a log eating some most delicious bugs. One jumped in the pool where its nice and cool, then there #{was_when} #{remaining_frogs} speckled frogs."
+
+    frog = frog.to_i - 1
+    remaining_frogs = remaining_frogs - 1
+
+  if frog.to_i == 2
```
francepack, line 33: Nice job catching this special case! Also well done making this work for any amount of frogs. Your next step with this challenge is to figure out how to translate your integers into spelled out numbers... there is no simple way outside of using a ruby gem, but key/value pairs could help you here.

day_7/caesar_cipher.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  cipher = normal.rotate(number.to_i)
+
+  ciphered = normal.zip(cipher).to_h
+
+  str.each_char.map { |c| p ciphered [c] }
+
+end
+
+cipher
```
francepack, line 35: You're on the right track here, but have you tried running this code? The first thing that happens is user input is obtained, THEN the first 2 questions get asked. You may play around with where your first p statement lives, or your could format this in a way where you collect the user input values first, then run method ` cipher` with parameters.

day_7/checker_board.rb
```diff
@@ -0,0 +1,39 @@
...
+elsif lines.to_i.even?
+  until lines.to_i == 0
+  p "X" + " " + "X" + " " + "X"
+  p " " + "X" + " " + "X" + " " + "X"
+  lines = lines.to_i - 2
+  end
+end
+end
+
+p checkerboard
```
francepack, line 39: Nice start here! I am noticing when I input an even number, I get the correct amount of rows, then a nil output, whereas when I try an odd number, I get one less row, but also a nil output, so there may be some logic to more closely examine. Also, a next step would be to make how long your rows are dynamic. Like when a user inputs 9, their rows will be "X X X X X".

francepack (commented):
Overall, excellent work on this! I'm glad to see you try both of the day 7 exercises. That type of work ethic leads to success at Turing- Keep pushing!

krislitman:
Thank you for taking the time to go over my work @francepack! It will be good for me to go back and work more on them as I've been improving my skills since completing these. I'll take some time this week to update my code and implement changes.

day_6/exercises/burrito.rb
krislitman: I went back through and tried using `@toppings << new` instead of `@toppings + new` and it worked!

PR #406, opened 2020-10-28
Doug Welchons

DougWelchons:
Doug Welchons Pre-Work

### DIFF
--- a/day_6/exercises/person.rb
+++ b/day_6/exercises/person.rb
@@ -1,5 +1,45 @@
-# Create a person class with at least 2 attributes and 2 behaviors. 
+# Create a person class with at least 2 attributes and 2 behaviors.
 # Call all person methods below the class and print results
 # to the terminal that show the methods in action.
 
-# YOUR CODE HERE
+class Person
+  attr_accessor :name, :age, :weight, :target_weight
+
+  def initialize (name, age, weight, target_weight)
+    @name = name
+    @age = age
+    @weight = weight
+    @target_weight = target_weight
+  end
+
+  def intro
+    puts "#{@name} is #{@age}. #{@name} currently weighs #{@weight}, and his target weight is #{@target_weight}"
+  end
+
+  def eat_junk_food
+    self.weight += 10
+    puts "#{self.name} now weighs #{self.weight}"
+  end
+
+  def exercise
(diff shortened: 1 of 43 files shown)


day_6/exercises/person.rb
```diff
@@ -1,5 +1,45 @@
...
+  end
+
+  def eat_junk_food
+    self.weight += 10
+    puts "#{self.name} now weighs #{self.weight}"
+  end
+
+  def exercise
+    if self.weight > self.target_weight
+      self.weight += -10
```
francepack, line 26: `-=` should also work.

day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,38 @@
...
+      puts "then there was #{self.number_of_frogs} speckled frog.\n\n"
+      self.rhyme
+    else
+      puts "#{self.number_of_frogs} speckled frog sat on a log \neating some most delicious bugs."
+      puts "One jumped in the pool where its nice and cool,"
+      puts "then there were no more speckled frogs!"
+    end
+  end
+end
+
```
francepack, line 32: Interesting strategy of calling rhyme again within that method until you are at the last verse! You could also explore using a loop here.

day_7/fizz_buzz.rb
```diff
@@ -0,0 +1,41 @@
...
+
+print "Please pick a starting number? "
+start = gets.chomp.to_i
+print "And and ending number? "
+ending = gets.chomp.to_i
+print "Fizz value? "
+fizz = gets.chomp.to_i
+print "buzz value? "
+buzz = gets.chomp.to_i
+
```
francepack, line 39: Nuanced approach here! Very clever to define values that will Fizz and Buzz! An OOP approach here (and with speckled frogs) may complicate things, but you've done well to get that practice in- the effort paid off as you have working, dynamic solutions!

francepack (commented):
Overall, excellent work on this! I'm glad to see you try both of the day 7 exercises- That type of work ethic leads to success at Turing! Keep up the hard work!

DougWelchons:
@francepack Thank you for the feed back. I will look into loops, and see how I could have implemented that instead of calling the rhyme again in the method.

PR #408, opened 2020-10-28
Wil McCauley

Wil-McC:
Submitting for review....a bunch of exercises!

### DIFF
--- a/day_7/10_speckled_frogs.rb
+++ b/day_7/10_speckled_frogs.rb
@@ -1,0 +1,28 @@
+def core_lines
+  puts "eating some most delicious bugs."
+  puts "One jumped in the pool, where it's nice and cool,"
+end
+
+#use hash for number word pairs
+# #{hash[num]} speckled frogs...
+
+
+def frog_rhyme(frogs)
+  counter = frogs
+
+  while counter > 2
+    puts "#{counter} speckled frogs sat on a log"
+    core_lines
+    counter -= 1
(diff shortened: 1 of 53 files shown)


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  puts "Two speckled frogs sat on a log"
+  core_lines
+  puts "Then there was one speckled frog."
+  puts "One speckled frog sat on a log"
+  core_lines
+  puts "Then there were no more speckled frogs!"
+end
+
```
francepack, line 27: Very nice solution here! I really like breaking out your core lines like that.

day_7/caesar_cipher.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+    end
+    # print output string to terminal
+    p encoded_str
+  end
+
+end
+
+cipher = CaesarCipher.new
+cipher.encode('Hotblack Desiato', 18)
```
francepack, line 40: Excellent work here. Nice use of an array to change your letters based on index. One tip would be that this could be modeled a bit more like a traditional ruby program, using gets.chomp to receive user input, then create an instance of your class with that input info.

francepack (commented):
Overall, excellent work on this! Keep pushing here- I think adding work to your day 7 challenges and trying both is great prep for Turing... that checkerboard is sneaky difficult! Excellent practice with breaking problems apart.

PR #409, opened 2020-10-28
Gaelyn Cooper

gaelyn:


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,54 @@
+# Print rhyme for 10 frogs AND Extension 2: Make it work for any number of frogs
+frogs = Array(1..10).reverse  # can change 10 to desired number of frogs
+
+frogs.each do |num|
+  if num == frogs[0]
+    puts "#{num} speckled frogs sat on a log"
+    puts "eating some most delicious bugs."
+    puts "One jumped in the pool where its nice and cool,"
```
francepack, line 8: It may be easier to divide by verse, rather than include the start of the next verse within your conditional. I see how you've divided it up, so you can just use num, but it may be more readable if you use `num - 1` from time to time.

day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,54 @@
...
+#   10 => "ten",
+#   9 => "nine",
+#   8 => "eight",
+#   7 => "seven",
+#   6 => "six",
+#   5 => "five",
+#   4 => "four",
+#   3 => "three",
+#   2 => "two",
+#   1 => "one"
```
francepack, line 34: Nice strategy to use printed words!

day_7/high_level.md
```diff
@@ -0,0 +1,32 @@
...
+This is the order in which I worked through how to solve this problem. I worked through it in an "inside out" manner.
+
+1. I know I need to be able to iterate through each letter of a string in order to shift the string so the first step is to find a way to convert the desired string into an array. I found `array = string.char` would accomplish this.
+2. Next I will need a block of code that will iterate through each character, rotate it `x` number of places in the alphabet, and then return the new array. This would be in the form of:
+```ruby
+array.each do |char|
+  #code that will rotate letters and return new array with new letters
+end
+```
+There are likely a number of different ways to accomplish this, but the method I came up with, and what seemed most logical, was to use a hash and make the keys letters and the values numbers. In this case, each `char` could be passed through the hash as a key which would correspond to a certain integer value. Then the character that was returned would be the character corresponding to the hash's integer value plus or minus `x`.
```
francepack, line 12: There are many ways to accomplish this. Also consider how arrays could help you keep your alphabet organized and accessible through index numbers.

francepack (commented):
Nice effort overall! I encourage you to keep working on the last challenges- working on that pseudocode, and research is great practice for project work.

PR #410, opened 2020-10-28
Andrew Smith

IcelandicGambit:
Here is my prework!


day_6/exercises/person.rb
```diff
@@ -1,5 +1,35 @@
...
+  end
+
+  def add_heel_lifts(new_height)
+    puts "adding heel lifts of #{new_height} cm"
+    @height = @height + new_height
+  end
+
+  def celebrate_birthday(new_age)
+    @age = @age + 1
+    puts "celebrating birthday, new age is #{@age}"
```
francepack, line 22: Since you do not use your parameter `new_age` inside of this method, you don't really need one! You can run this method without needing a parameter... seems like it will add 1 to an instances age no matter what their current age is.

day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,18 @@
...
+  puts "#{number_of_frogs} speckled frogs sat on a log"
+  number_of_frogs = number_of_frogs - 1
+  puts "eating some most delicious bugs."
+  puts "One jumped in the pool where its nice and cool,"
+
+
+  if number_of_frogs == 0
+    puts "then there were no more speckled frogs!"
+  elsif number_of_frogs > 1
+    puts "then there were #{number_of_frogs} speckled frogs."
```
francepack, line 14: Very nice conditional! You've done well to see that order matters here, and kept things pretty simple. I like just tracking the one variable, `number_of_frogs`, and you leverage WHEN you modify that variable to use it in different ways.

francepack (commented):
Overall, nice work here. I am missing complete work for day 7- we've asked you to pseudocode out either the cipher or the checkerboard, and I do not see that work here. Please send me a slack message when you are able to complete day 7. I highly encourage not just writing pseudocode, but then trying to code some of this out, following steps you have laid out for yourself.

day_7/high_level.rb
```diff
@@ -0,0 +1,30 @@
...
+        if c == " "
+          output = output + " "
+        else
+          if c.upcase == c
+            output = output + upper_case[(c.ord - 65 + shift_amount) % 26]  #c.ord gets you the position of the character in the ascii array, subtracting 65 takes the "A" position and converts it into the position in the upper_case array. Modulo limits the range of outputs to the number of letters in the alphabet, so the code will "wrap around" if I make the shift longer, for example.
+          elsif c.downcase == c
+            output = output + lower_case[(c.ord - 97 + shift_amount) % 26]
+          end
+        end
+      end
```
francepack, line 22: This is worth some examination- I'm sure you have, but spending some time really analyzing what code block each `end` closes may help you keep organized... why does your return output need to be where it is? If you don't know the answer, that is ok, but a question to hold onto as you learn more about context and scope.

day_7/high_level.rb
```diff
@@ -0,0 +1,30 @@
...
+      output  #This syntax had to be explained to me, the ends and where to place output.
+    end
+end
+
+#Test code
```
francepack, line 27: I really like the OOP approach! Something you could tinker with in the coming weeks here, and with fizzbuzz, is to leverage gets.chomp to gather user input, and run your programs based on what the user enters.

PR #412, opened 2020-10-29
Jacob Arellano

jakejakearell:
Here is my prework!

### DIFF
--- a/day_7/projects/10_speckled_frogs.rb
+++ b/day_7/projects/10_speckled_frogs.rb
@@ -1,0 +1,51 @@
+frogs = [10, 9, 8, 7, 6, 5, 4, 3, 2, 1]
+
+frogs.each do |frog|
+  if frog == 1
+    puts "1 speckled frog sat on a log eating some most delicious bugs.
+    It jumped in the pool where its nice and cool, then there were no more speckled frogs!"
+  else
+    puts "#{frog} speckled frogs sat on a log eating some most delicious bugs.
+        One jumped in the pool where its nice and cool, then there was #{frog -1} speckled frogs."
+  end
+end
+#Extension 2
+
+frogs_string = {
+  10 => 'ten',
+  9 => 'nine',
+  8 =>'eight',
+  7 => 'seven',
+  6 => 'six',
+  5 => 'five',
+  4 => 'four',
+  3 => 'three',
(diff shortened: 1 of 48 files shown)


day_7/projects/10_speckled_frogs.rb
```diff
@@ -0,0 +1,51 @@
+frogs = [10, 9, 8, 7, 6, 5, 4, 3, 2, 1]
```
mschae16, line 1: @jakejakearell - While testing out this code, the rhyme correctly prints out counting down from 10, however I noticed a couple issues, specifically around wording - for example, the first iteration prints out:

> 10 speckled frogs sat on a log eating some most delicious bugs.  One jumped in the pool where its nice and cool, then there was 9 speckled frogs.

The issue here is that it should say "there **were** 9 speckled frogs".  The same issue happens again throughout.  When we get to the second to last iteration, it prints out `"then there was 1 speckled frogs."` Since there's only 1 left, this should actually say "frog" singular.

How could you update this exercise to correctly print out each iteration of the rhyme?

day_7/projects/10_speckled_frogs.rb
```diff
@@ -0,0 +1,51 @@
...
+    puts "#{string.capitalize} speckled frogs sat on a log eating some most delicious bugs.
+     One jumped in the pool where its nice and cool, then there was #{frogs_string[frog - 1]} speckled frogs."
+  end
+end
+
+
+
+#Extension 3
+#
+frogs = (1..3).to_a.reverse #could change  the 3 into any interger and program woudl work
```
mschae16: Nice work tackling both extensions.  For this third extension, I think the idea is actually to make the number of frogs to print dynamic, so that a user could input any number and it would print the appropriate iterations of the rhyme.  Can you think of a way to make it dynamic?

mschae16 (commented):
@jakejakearell - Nice work on this pre-work. Glad to see you went above and beyond and tackled those extensions.  One note on your commits - I noticed you used the message `Initial commit` many times, which is not generally seen since you would only ever use that message for your **very first commit** in a repository.  Commit messages should tell the reader (who may be yourself in the future) what was updated or added, and the message "Initial commit" doesn't really tell us anything about what changed.

PR #415, opened 2020-11-02
Jenny Branham

jbranham1:


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,44 @@
...
+    if number > 2
+      puts rhyme + rhyme_ending
+    elsif number ==2
+      rhyme_two_ending = rhyme_ending.gsub("frogs", "frog").gsub("were", "was")
+      puts rhyme + rhyme_two_ending
+      # puts "#{number} speckled frogs sat on a log eating some most delicious bugs."
+      # puts "One jumped in the pool where its nice and cool,"
+      # puts "then there was #{x} speckled frog.\n\n"
+    else
+      rhyme_one = rhyme.gsub("frogs","frog")
```
francepack, line 25: Nice strategy to avoid 'hard coding' those lines out again!

day_7/checker_board.rb
```diff
@@ -0,0 +1,23 @@
...
+end
+
+def even_row
+  return "X X X"
+end
+
+def odd_row
+  return " X X X"
+end
+
```
francepack, line 18: I'm liking your dev instinct to break things into multiple single purpose functions! For this challenge, the next step will be making your `even_row` and `odd_row` functions responsive to the the amount of Xs and " "s you have, so if your row input was 2, you'd just have 'X ' or ' X'

day_7/to_words/to_words.gemspec
```diff
@@ -0,0 +1,28 @@
+require_relative 'lib/to_words/version'
```
francepack, line 1: I'm unsure how much config you needed to do to get a gem up and running, but I'm sure that was a learning experience into itself! Another challenge is to not use a gem, and see how key/value pairs can serve as a reference point to change integers into spelled out words.

francepack (commented):
Overall, excellent work on this! I'm glad to see you try both of the day 7 exercises. That type of work ethic leads to success at Turing- keep up the hard work!

PR #418, opened 2020-11-03
Tommy Nieuwenhuis

tsnieuwen:


day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,44 @@
...
+  else
+    was_were = "was"
+  end
+
+  if new_frog_number > 1 || new_frog_number == 0
+    new_frog_plural = "frogs"
+  else
+    new_frog_plural = "frog"
+  end
+
```
francepack, line 15: You can probably combine your first and second conditionals, then your 3rd ands 4th ones. You can set 2 different variables within an if block.

day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,44 @@
...
+    jumper = "It"
+  end
+
+  puts "#{x} speckled #{frog_plural} sat on a log
+  eating some most delicious bugs.
+  #{jumper} jumped in the pool where it's nice and cool,
+  then there #{was_were} #{new_frog_number} speckled #{new_frog_plural}."
+
+end
+
```
francepack, line 34: Nice work just having 1 area where you type the whole poem out! It can be easy here to go down a path where you type this out in multiple locations in your code.

day_7/checker_board.rb
```diff
@@ -0,0 +1,20 @@
+size_of_board = 10
+
```
francepack, line 2: A next step with some of these exercises could be to leverage gets.chomp to run these based on user input.

day_7/high_level.md
```diff
@@ -0,0 +1,57 @@
...
+    if checker % 2 == 0
+      puts " ""X" * (size_of_board / 2)
+    elsif checker % 2 != 0
+      puts "X"" " * (size_of_board / 2) + "X"
+    end
+  end
+end
+
+```
+
```
francepack, line 55: Excellent pseudocode! I hope you found this helped you keep your thoughts organized.

francepack (commented):
Overall, very nice work here! I suggest turning some of the day 7 challenges into little ruby programs that grab some user input, and run based on that. Also, I think you knocked out the harder challenge with the checkerboard, so doing the cipher will be excellent practice leading up to Turing as well. Keep up the hard work!

PR #420, opened 2020-11-04
Khoa Nguyen

Omegaeye:
Here is my back end prework for 2011 cohort.

Link to my technical: https://github.com/Omegaeye/backend_mod_1_prework


day_6/exercises/burrito.rb
```diff
@@ -1,16 +1,30 @@
...
 
 class Burrito
   attr_reader :protein, :base, :toppings
+
   def initialize(protein, base, toppings)
     @protein  = protein
     @base     = base
     @toppings = toppings
   end
+
```
francepack, line 15: Where is your logic here? This is a good opportunity to do a bit more research on common ways to add something to, or remove something from an array.

day_6/questions.md
```diff
@@ -1,13 +1,42 @@
...
+  end
+
+  def sniff
+    p "It smelled something funny."
+  end
+
+  def sit
+    @hungry = true
+  end
+
```
francepack, line 32: While I think just about any action would make a dog hungry, (when AREN'T they hungry? hah) this relationship is not as intuitive to me... I would think something more active would be more likely to make the dog hungry.

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,14 @@
+(1..100).map do |x|
+ f3 = x.modulo(3) == 0
+ f5 = x.modulo(5) == 0
+
```
francepack, line 4: I don't often see the .modulo method used... more common is the modulo operator which is just 1 key. Can you remember which that is? If not, I bet some quick googling can tell you!

day_7/ceasar_cipher.rb
```diff
@@ -0,0 +1,30 @@
...
+class CeasarCipher
+
+  def encode(string, key)
+    convert_string = string.chars.map {|char| char.ord}
+    shifted = convert_string.map {|char| char+key}
+    shifted.map { |char| char.chr }.join
+    end
+end
+cipher = CeasarCipher.new
+p cipher.encode("Hello World", 5)
```
francepack, line 30: Good start here! character keys can be difficult to work with because there are keys that are not letters. For instance, when I run your code, the output for 'hello world' shifted 5 characters is "Mjqqt%\\twqi". The next step will be figuring out strategies to not shift and empty spaces, as well as to keep the key codes within the range of letter characters.

day_6/exercises/burrito.rb
Omegaeye: @francepack I don't know what happened there. I have re-written the code below. I also have edit my files in my repo. Thank you for your pointers.

```ruby
class Burrito
  attr_reader :protein, :base, :toppings

  def initialize(p, b, t)
    @protein  = p
    @base     = b
    @toppings = t
  end

  def add_toppings(topping)
    @toppings.push(topping)
    puts "For sure, I can add #{topping} for you"
  end

  def remove_toppings(rt)
    @toppings.delete(rt)
    p "Not a problem at all, we can remove #{rt} for you."
  end

  def change_protein(new_protein)
    @protein.replace(new_protein)
    p "Load them #{new_protein} up!"
  end

end

p dinner.protein
p dinner.base
p dinner.toppings
dinner.add_toppings("pepperoni")
puts "So you have #{dinner.toppings} toppings in your burrito."
dinner.remove_toppings("salsa")
puts "Now you have #{dinner.toppings}."
dinner.change_protein("Chicken")
puts "So you have picked #{dinner.protein}, #{dinner.base}, and #{dinner.toppings} in your burrito."
```

day_6/questions.md
Omegaeye: @francepack I was thinking along the line of that when my dog is hungry or want something. They need to sit first.

day_7/ceasar_cipher.rb
Omegaeye: @francepack I was looking into that. The code I had above that was greyed out worked that way. It doesn't convert the space into a character and I can also not shift it.

day_7/fizzbuzz.rb
Omegaeye: @francepack Thank you for the tip. Please see below.

```ruby
 f3 = x % 3 == 0
 f5 = x % 5 == 0
```

PR #423, opened 2020-11-04
Trevor Suter

Trevorsuter:
Here is my prework!!


day_6/exercises/person.rb
```diff
@@ -1,5 +1,31 @@
...
+  end
+
+  def change_age(a)
+    age.replace(a)
+  end
+
+  def change_name(first)
+    name.replace(first)
+  end
+
```
francepack, line 19: I don't see the `replace` method used all that much in these scenarios. I believe it works fine, but you can also just reassign your attributes.

day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,38 @@
...
+    if (enum) == 0
+      puts "no more"
+  end
+
+  def info
+    "#{number}"
+  end
+end
+end
+
```
francepack, line 18: You've got a lot of `puts` statements you are leaning on here! I would steer away from an OOP solution here. It can work, but it can over complicate this one. It could be good practice to redo this one, potentially getting some user input to inform how many times a loop should be run. This way, you can just type out repeated verses once, and your program will be more dynamic.

day_7/high_level.md
```diff
@@ -0,0 +1,13 @@
+### Caesar Cipher ###
+
+#####I chose to try and work through building a caesar cipher. Here is a step-by-step of my process:#####
+
+1. First thing to do is read the instructions given to me thoroughly. Sometimes, the answer to your simplest problems come right from the source, so I *always* keep the instructions ready.
+
```
francepack, line 6: Yessir!

day_7/high_level.md
```diff
@@ -0,0 +1,13 @@
...
+2. Next thing I do is try and work through the problem myself. Theres a lot of methods that I don't quite know yet, but if I can work through with what I already know, I tend to create a better base for when I *do* need to start querying google.
+
+3. Next, I go to google and search 'caesar cipher ruby' and read through trusted websites, such as ruby-docs and stack overflow. If I can find a sample code for the caesar cipher, I'll have a better understanding of the methods used.
+
```
francepack, line 10: Looking up code for these challenges can help, but I also think you have some tools already for breaking this problem apart. When you start thinking about these kinds of things, I often start with some simple questions like 'what am I going to put into this method?' and 'what do I want the method to return to me?' In this case, you're going to give it a string, and then return a string. Even such simple answers to questions can give you starting points to research methods. For instance, since you want to do something to each individual letter, what might you want to do to your string data so you can iterate over each letter?

Keep practicing breaking these problems apart! It certainly takes practice and patience to make all your steps, then research them, but I think breaking the problem down into much smaller steps is essential to keeping your research organized.

francepack (commented):
Overall, nice work here! Keep up the practice all around: I highly recommend the mental exercise of creating classes out of objects around you, or on a walk, thinking of what you'd name attributes and methods. I also recommend approaching the checkerboard and cipher pseudocode in a bit more detail- what are the specific things you need to do to accomplish that challenge? Like 1. get string data 2. split this data so I can look at each letter ... so on and so forth. Keep up the hard work!

PR #425, opened 2020-11-04
Harrison Blake

harrison-blake:
This is my pre work

### DIFF
--- a/day_6/exercises/burrito.rb
+++ b/day_6/exercises/burrito.rb
@@ -1,4 +1,4 @@
-# Add the following methods to this burrito class and 
+# Add the following methods to this burrito class and
 # call the methods below the class:
 # 1. add_topping
 # 2. remove_topping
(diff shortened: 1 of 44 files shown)


day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,32 @@ def initialize(protein, base, toppings)
...
   end
+
+  def add_topping(topping)
+    self.toppings.push(topping)
+  end
+
+  def remove_topping(r_topping)
+    self.toppings.each do |topping|
+      if topping == r_topping
+        toppings.delete(topping)
```
francepack, line 22: Very nice use of a loop to examine each element of your toppings array!

day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  else
+    puts "1 speckled frog sat on a log
+    eating some most delicious bugs.
+    One jumped in the pool where its nice and cool,
+    then there were no more speckled frogs!"
+    break
+  end
+  count -= 1
+end
```
francepack, line 18: Good start here- When your count is 2, there is also a special case because of the pesky phrasing in the last line (was 1 more, not were 1 one more). Also, try to explore how you may leverage gets.chomp to gain user input to set your count at different numbers.

day_7/checker_board.rb
```diff
@@ -0,0 +1,18 @@
...
+    elsif col % 2 == 1 && row % 2 == 1
+      print "X"
+    else
+      print " "
+    end
+    col += 1
+  end
+  row += 1
+  print "\n"
+end
```
francepack, line 18: Excellent results here! Again, I think you can explore getting some user input. I'm also wondering whether you need to create an array to iterate over every element of that array, or if you could just do something a certain amount of times... I think you could avoid the array creation, but perhaps there is something I'm missing.

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+  if count % 5 == 0
+    printed += "Buzz"
+  end
+  if count % 3 != 0 && count % 5 != 0
+    printed = count
+  end
+    puts printed
+    count+=1
+end
```
francepack, line 22: Nice job constructing an output! Very useful pattern to get the hang of.

francepack (commented):
Overall, excellent work on this! I'm glad to see you try both of the day 7 exercises. That type of work ethic leads to success at Turing- keep up the hard work!

PR #427, opened 2020-11-15
Joe Mecha

joemecha:
Mod 1 prework completed!


day_6/exercises/burrito.rb
```diff
@@ -11,9 +11,29 @@ def initialize(protein, base, toppings)
     @base     = base
     @toppings = toppings
   end
+
+  def add_topping(topping)
+    @toppings << topping
+  end
+
```
francepack, line 18: Nice attribute/method interactions here!

day_7/10_speckled_frogs.rb
```diff
@@ -0,0 +1,25 @@
...
+      puts puts "#{frog_qty} speckled frog sat on a log eating some most delicious bugs.\n
+One jumped in the pool where it's nice and cool, then there were no more speckled frogs!"
+      frog_qty -= 1
+    else
+      break
+    end
+  end
+end
+
+ten_speckled_frogs(10)
```
francepack, line 25: Nice job seeing all the conditions to hit here. A next step could be to model this like a little ruby program, gathering some user input to run in your function.

day_7/caesar_cipher.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+  puts encoded_array.join
+end
+
+puts "Caesar Cipher"
+puts "-------------"
+print "Please enter the word(s) you want to encode: "
+string = gets.chomp
+print "Enter the number of characters to shift by (ex. 3 shifts A to D; -3 shifts A to X): "
+shift_value = gets.chomp.to_i
```
francepack, line 30: Nice job leveraging user input here! Also, way to stick with this challenge and push through! Seeing loots of good progress across the different cipher files. Some of this stuff- you can just find the answers out there, but grappling with it, breaking code, and experimentation are the best ways to learn the ins and outs- very well done here.

day_7/caesar_cipher_attempt1.rb
```diff
@@ -0,0 +1,30 @@
+# This program takes a user supplied string and shift value to return a
+# shift cipher version of the string in upper case characters
+
+def encode
+  cipher_hash = {"A" => 0, "B" => 1, "C" => 2, "D" => 3, "E" => 4, "F" => 5 "G" => 6, "H" => 7, "I" => 8, "J" => 9, "K" => 10, "L" => 11, "M" => 12, "N" => 13, "O" => 14, "P" => 15, "Q" => 16, "R" => 17, "S" => 18, "T" => 19, "U" => 20, "V" => 21, "W" => 22, "X" => 23, "Y" => 24, "Z" => 25}
+  cipher_array = ["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z"]
```
francepack, line 6: arrays are just hashes where the 'key' is the index number, and the element is the value. If that doesn't make sense right way, just something to stew on.

day_7/checker_board.rb
```diff
@@ -0,0 +1,41 @@
...
+          print "X"
+        end
+        square_counter -= 1
+      end
+    end
+    puts ""
+    row_counter -= 1
+  end
+end
+
```
francepack, line 29: Way to stay organized! Loops within loops can get hairy.

day_7/fizzbuzz.rb
```diff
@@ -0,0 +1,25 @@
...
+  # Create array of integers 1 to n
+  num_array = (1..n).to_a
+  # Iterate through the elements of the array
+  num_array.each do |a|
+  # For any number that is a multiple of both 3 and 5, print 'FizzBuzz'
+    if (a % 3) == 0 && (a % 5) == 0
+      print "FizzBuzz, "
+# For any number that is a multiple of 3, print 'Fizz'
+    elsif (a % 3) == 0 && (a % 5) != 0
+      print "Fizz, "
```
francepack, line 14: You don't need the second part of the conditionals here- if both applied, it'd be swooped up by your first condition! You can just run the second 2 conditions checking for the 0 value of a modulo, rather than needing to see if it != 0.

francepack (commented):
Overall, strong work and effort! Very nice job working on the challenges at the end- all the effort there will pay off at Turing, being familiar with code challenges, breaking those problems apart, and digging in to research. Keep up the practice and experimentation!

day_7/caesar_cipher.rb
joemecha: Mason -- thanks for the encouragement here, and for reviewing my all of my work -- cheers!

PR #445, opened 2021-01-13
Alexander Brueck

brueck1988:
Here is my prework!

### DIFF
--- a/day_7/projects/10_speckled_frogs.rb
+++ b/day_7/projects/10_speckled_frogs.rb
@@ -1,0 +1,73 @@
+def in_words(int)
+  numbers_to_name = {
+      1000000 => "million",
+      1000 => "thousand",
+      100 => "hundred",
+      90 => "ninety",
+      80 => "eighty",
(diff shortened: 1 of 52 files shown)


day_7/projects/10_speckled_frogs.rb
```diff
@@ -0,0 +1,73 @@
...
+      return str + "#{name.capitalize}" if int%num == 0
+      return str + "#{name.capitalize} " + in_words(int%num)
+    elsif int/num > 0
+      return str + in_words(int/num) + " #{name.capitalize} " + in_words(int%num)
+    end
+  end
+end
+
+
+
```
francepack, line 50: Way to tackle the nuances of this challenge! I like how you've set this up like a program, gathering user input to set the range.

francepack (commented):
Very nice work here- I don't have much actionable feedback as it looks like you've worked with those challenges quite a bit. Remember as you start your Truing project work to start with your pseudocode, and go from there. Keep up the hard work!

